Issue #119, opened 2025-07-10
Webhook Listener for PP Volume Releases


linkion:
### Feature

When a PP Volume is released, the bot will:
- Make a post with the Title of the Volume and volume number.
- Titles of the Articles in the Volume.
- Link to the PP Volume on the PP website.

tadjh:
Sunsetting monorepo issues

Issue #76, opened 2025-07-06
Discord webhook setup for volume publications


linkion:
When new volumes are published send out a notification via discord webhook. Could be used to easily integrate Pragmatic Papers publication announcements on discord server like DGG and DGG Political Action

tkmalmberg:
Related to https://github.com/dggpoliticalaction/discord-bot/issues/2

PR #153, opened 2025-07-23
Added webhooks to Pragmatic Papers

lith-x:
notes: as of now, the key `fetch` call is aimed at the [discord api](https://discord.com/developers/docs/resources/webhook#execute-webhook). avatar URL could probably be moved to a more accessible spot, the check for whether or not to actually push a webhook event could probably be more sophisticated. at the moment it's a volume going draft -> published while having no previous publication date (i.e. the first publish)

this *just* pushes the raw URL, nothing fancy. having a bot grab it and do fancy stuff is probably the play, as of right now you could "add webhook" in the discord server and the message would be the URL (with a username / avatar override)

partially addresses #119 / #76

### DIFF
--- a/apps/pragmatic-papers/src/collections/Volumes/hooks/pushToBot.ts
+++ b/apps/pragmatic-papers/src/collections/Volumes/hooks/pushToBot.ts
@@ -34,7 +34,11 @@ export const pushToBot: CollectionAfterChangeHook<Volume> = async (args) => {
       console.error(e)
     })
 
-    if (!res || !res.ok) return
+    if (!res || !res.ok) {
+      console.error(`Request to webhook ${webhook.name} failed`)
+      if (res) console.error(`    ${res.status}: ${res.statusText}`)
+      return
+    }
     const articlesPushed = webhook.pushed ?? []
     articlesPushed.push({ volumeNumber: args.doc.volumeNumber, id: args.doc.id.toString() })
 
--- a/apps/pragmatic-papers/src/collections/Volumes/hooks/pushToBot.ts
+++ b/apps/pragmatic-papers/src/collections/Volumes/hooks/pushToBot.ts
@@ -1,9 +1,10 @@
 import { type Volume } from '@/payload-types'
 import { getPayload, type CollectionAfterChangeHook } from 'payload'
 import config from '@payload-config'
+import { env } from 'process'
 
 export const pushToBot: CollectionAfterChangeHook<Volume> = async (args) => {
-  // TODO: unpublishing then republishing will not trigger a webhook push
+  // NOTE: unpublishing then republishing will never trigger a webhook event
   if (
     args.previousDoc._status != 'draft' ||
     args.doc._status != 'published' ||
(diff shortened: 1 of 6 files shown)


apps/pragmatic-papers/src/collections/Volumes/hooks/pushToBot.ts
```diff
@@ -0,0 +1,53 @@
...
+    const articlesPushed = webhook.pushed ?? []
+    articlesPushed.push({ volumeNumber: args.doc.volumeNumber, id: args.doc.id.toString() })
+
+    payload.update({
+      collection: 'webhooks',
+      id: webhook.id,
+      data: {
+        pushed: articlesPushed,
+      },
+    })
```
linkion, line 51: [nitpick] We should name `articlesPushed` to `volumesPushed`, since that's what we are pushing for

apps/pragmatic-papers/src/collections/Webhooks/index.ts
```diff
@@ -0,0 +1,66 @@
...
+            readOnly: true,
+            hidden: true,
+          },
+        },
+      ],
+      admin: {
+        readOnly: true,
+        hidden: true,
+      },
+    },
```
linkion, line 64: We should also store like a `lastPushedAt` dateTime, just for debugging

apps/pragmatic-papers/src/collections/Webhooks/index.ts
```diff
@@ -0,0 +1,66 @@
...
+            return (
+              ctx.data?.pushed
+                ?.map((v) => v.volumeNumber ?? 0)
+                .reduce((prev, curr) => (prev > curr ? prev : curr))
+                ?.toString() ?? '-'
+            )
+          },
+        ],
+      },
+    },
```
linkion, line 42: Probably should make read only
```suggestion
    {
      type: 'text',
      name: 'mostRecentPushed',
      label: 'Most Recent',
      hooks: {
        afterRead: [
          (
            ctx: FieldHookArgs<Webhook, unknown, { pushed: { volumeNumber: number }[] }>,
          ): string => {
            return (
              ctx.data?.pushed
                ?.map((v) => v.volumeNumber ?? 0)
                .reduce((prev, curr) => (prev > curr ? prev : curr))
                ?.toString() ?? '-'
            )
          },
        ],
      },
      admin: {
        readOnly: true,
        description: 'The most recent volume number that has been pushed to this webhook',
      },
    },
```

linkion (changes requested):
Requesting some changes, but overall looks good.

apps/pragmatic-papers/src/collections/Volumes/hooks/pushToBot.ts
lith-x: slipped that articles != volumes, very explicitly so lmao, changed that

apps/pragmatic-papers/src/collections/Webhooks/index.ts
lith-x: implemented the datetime that the volume was pushed, using that to determine "latest", displaying it in the column, yeehaw

PR #164, opened 2025-08-06
Fix/rules updates

MBasala:
Updated the rules file with the new rules strings. Make minor changes to package.json such as author and package name.


linkion (approved):
LGTM

PR #166, opened 2025-08-09
Update dgg website template

linkion:
Updated the dgg-political-action app to match the new template from Payload that includes setup configs for testing, and new sessions


linkion:
Stale

PR #170, opened 2025-08-16
remove update-server-count-job from start-manager.ts, remove template owner stream and paypal links

tkmalmberg:
This function sets the bot's activity to "Streaming to ${Server_count} servers"

We also don't plan on putting this bot on any other servers so there's no need for this job to run.

closes https://github.com/dggpoliticalaction/org-repo/issues/161

### DIFF
--- a/apps/discord-bot/lang/lang.common.json
+++ b/apps/discord-bot/lang/lang.common.json
@@ -25,7 +25,7 @@
   "links": {
     "author": "https://github.com/",
     "docs": "https://top.gg/",
-    "donate": "",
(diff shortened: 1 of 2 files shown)


linkion (changes requested):
If you can fix the unused var issue, I'll begin to review more closely

linkion:
> If you can fix the unused var issue, I'll begin to review more closely

lol, nvm

linkion (approved):
Tested, appears to be working as expected, the streaming activity is no longer appearing

Image

PR #176, opened 2025-08-21
Moved Footer and Header to globals/ folder

r29vzg:
- Just a small cleanup PR that better organizes the project folder structure, as more Globals in this project grows, it will start to clutter the `src` folder. This simply moves the Globals into a  `src/globals` folder, similar to `collections/`,  `heros/`, and `blocks/`


linkion:
> > Tested, I agree with having these components in a global folder
> 
> I suppose it's already done but I think I disagree with this change. I was pretty suspicious of heros and blocks being outside of src to begin with.
> 
> On a slightly different point, "globals" feels unclear to me. As a contributor, what dictates whether a component should go in src/components or globals?

Imo, things that appear on every page is a global, like out Pragmatic Papers header and the Footer, idk, I think it's okay to move these from the src into a globals

And then components are our individual components of our website, like the media component and pagination

TedTimbrell:
> > > Tested, I agree with having these components in a global folder
> 
> > 
> 
> > I suppose it's already done but I think I disagree with this change. I was pretty suspicious of heros and blocks being outside of src to begin with.
> 
> > 
> 
> > On a slightly different point, "globals" feels unclear to me. As a contributor, what dictates whether a component should go in src/components or globals?
> 
> 
> 
> Imo, things that appear on every page is a global, like out Pragmatic Papers header and the Footer, idk, I think it's okay to move these from the src into a globals, also global constants later
> 
> 
> 
> And then components are our individual components of our website, like the media component and pagination


I feel like we're describing the current usage of a component rather than what the component does or is. If the header ends up getting disabled on some percentage of pages because we're doing a fullscreen activity is it still global?

r29vzg:
> > > > Tested, I agree with having these components in a global folder
> > 
> > 
> > > 
> > 
> > 
> > > I suppose it's already done but I think I disagree with this change. I was pretty suspicious of heros and blocks being outside of src to begin with.
> > 
> > 
> > > 
> > 
> > 
> > > On a slightly different point, "globals" feels unclear to me. As a contributor, what dictates whether a component should go in src/components or globals?
> > 
> > 
> > Imo, things that appear on every page is a global, like out Pragmatic Papers header and the Footer, idk, I think it's okay to move these from the src into a globals, also global constants later
> > And then components are our individual components of our website, like the media component and pagination
> 
> I feel like we're describing the current usage of a component rather than what the component does or is. If the header ends up getting disabled on some percentage of pages because we're doing a fullscreen activity is it still global?

According to the documentation, a Global is basically a singleton that should only appear one time on a page:
https://payloadcms.com/docs/configuration/globals

So that means in the future if we build out something like a "Activity Feed" for example, that would be contained within a Global or something like that.

PR #177, opened 2025-08-21
Updates base squiggle component with size prop

r29vzg:
#### Squiggle Sizing
Moves the size prop which was previously just on the Block, to the actual base component, to allow for consistent sizing of the squiggle component regardless of where it's used, inside of blocks or inside of sub-components.

In this case, the Squiggle component was used within the `VolumesView` component as a divider between the volumes, pervious,  before this change it the `max-width` was constrained by it's parent component meaning it would be different wherever used unless properly sized.

#### Squiggle moved to index for volumes
Instead of rendering the squiggle in the Volumes View Entry component, I moved it to the index, so we can hide it from the last one a bit easier.


linkion (approved):
Tested, this looks good

PR #181, opened 2025-08-21
Refactor header footer

linkion:
Reopened #176 to resolve discussions before pulling into codebase


linkion:
Here’s a summary of the arguments from the pull request discussion about moving the **Header** and **Footer** into a `globals/` folder:

**Arguments in Favor (r29vzg, linkion):**

* **Organization & Cleanup:** As more globals are added, keeping them in the root `src/` would clutter the folder. Moving them to `src/globals/` is consistent with how `collections/`, `heros/`, and `blocks/` are organized.
* **Definition of Globals:** Components that appear across every page (like Header and Footer) are considered "globals." Later, global constants or features (like an activity feed or call-to-action banner) could also live there.
* **PayloadCMS Alignment:** According to PayloadCMS docs, a *Global* is a singleton that should only appear once per page. This makes Header and Footer fit naturally in `globals/`.
* **Scalability:** As the article and block system grows more complex, a proper folder structure now will prevent messiness later.

**Arguments Against / Concerns (TedTimbrell):**

* **Unclear Distinction:** The term *globals* feels vague—contributors may not know when to put something in `src/components/` vs. `globals/`.
* **Current Usage vs. Identity:** The categorization seems based on *how a component is used* (appears everywhere) rather than *what it is*. If the Header were disabled on some pages (e.g., fullscreen activity), does it stop being a global?
* **Suspicion of Structure Drift:** Skepticism about following the same path as `heros/` and `blocks/` being outside of `src/`, which might fragment project structure.

**Outcome:**

* The PR was initially merged, but soon after, **linkion reverted the change** (#180), likely due to the conceptual confusion around what counts as a “global.”
* A follow-up refactor PR (#181) was opened to revisit how Header and Footer should be organized.

👉 In short:

* **Pro side:** Organization, PayloadCMS consistency, scaling needs.
* **Con side:** Conceptual vagueness, potential misuse, structural fragmentation.

r29vzg:
Outlined here: https://github.com/dggpoliticalaction/org-repo/pull/176#issuecomment-3211167524

Given that it's Globals is a domain-specific PayloadCMS term, I think that moving it to a `globals/` folder to be aligned with the other folders is as good idea.

Also, I think it's somewhat important to remember that a lot of the conventions that we have on this project so far just come from the PayloadCMS with Next.js website starter found here:
https://github.com/payloadcms/payload/tree/main/templates/with-vercel-website

The globals in `src/` comes straight from there. But I really advocate us moving them into it's own folder.

linkion:
> @linkion @TedTimbrell 
> 
> What are we thinking? Can we get this merged in?

I'll review this tonight

TedTimbrell:
It just seems off to me, from reading the [doc page](https://payloadcms.com/docs/configuration/globals) and the [one example that uses it](https://github.com/payloadcms/payload/blob/main/examples/live-preview/src/globals/MainMenu.ts) it feels like a place for configs rather than the dumb components themselves?

If we want to move these to say, `src/navigation` and be really opinionated that `src/components` should contain next to no business logic, that's fine by me?

r29vzg:
It's actually for both. In practice, the way that Payload works is in two parts, you have a config that describes the shape of the data which is used to build the Typescript types and the database schema, as well as construct the API payload for the frontend. Then there is a payload view part or the rendering part, paired with the data described in the config, you need to now render the described data. In practice it makes a lot of sense to put both the config and the component together in the same folder.

This pattern is used in the blocks folder already, which is used to render specific types of `Collections` in PayloadCMS, the difference is that collections can be rendered in many ways, and Collections are made up of difference Blocks. A `Global` as described by PayloadCMS is just a singleton, so it should generally naturally only one have way other render it, however that's not always the case, but it seems to be that way in proactive.

PR #185, opened 2025-08-24
CTA Post Reaction Reading

tkmalmberg:
This feature will be used to monitor the completion rate of CTAs per region.

Making this a draft for now to work out some of the timing details. I might be able to make it update every hour our so as long as the  existing thread is still active.


tkmalmberg:
closes https://github.com/dggpoliticalaction/org-repo/issues/120

tkmalmberg:
Code completed and ready for review.

linkion (changes requested):
Some things that need changing:
- Restarting reaction collecting
  - We could probably look back 2 weeks for any CTA messages on start to create threads
  - Then look for existing threads back 6 months for updating
- Live update of chart, then check on interval

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  private async activeCTAThread(msg: Message): Promise<PublicThreadChannel | undefined> {
+    if (msg.hasThread && msg.thread?.type === ChannelType.PublicThread && !msg.thread?.archived) {
+      return msg.thread;
+    }
+  }
+
+  private async createCTAThread(msg: Message): Promise<PublicThreadChannel | undefined> {
+
+    if (msg.channel.type === ChannelType.GuildText) {
```
prophetofxenu, line 110: This is more of a personal preference, but `msg` should always be `ChannelType.GuildText` since the calling function already does that check. Is there a TypeScript error if you don't do this check? If so, I personally prefer either throwing an exception on the invalid condition or making a type assertion. Returning `undefined` in conditions like this can make it hard to determine where logic is falling apart.

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+  private async startCTAReactionCollector(msg: Message, thread: PublicThreadChannel): Promise<void> {
+    const filter = (reaction, user) => reaction.emoji.name === finishedEmoji && user.id != msg.client.user.id
+    const collector = msg.createReactionCollector({ filter, time: oneHour });
+    const roleReactions = await this.getFinishedReactions(msg)
+
+    collector.on('collect', async (reaction, user) => {
+      const member = msg.guild?.members.fetch(user.id)
+      if (member != undefined) {
+
+        console.log(`Got ${reaction.emoji.name} from ${(await member).displayName}`)
```
prophetofxenu, line 142: Another candidate for a debug log.

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  // start a reaction collector on the given message
+  private async startCTAReactionCollector(msg: Message, thread: PublicThreadChannel): Promise<void> {
+    const filter = (reaction, user) => reaction.emoji.name === finishedEmoji && user.id != msg.client.user.id
+    const collector = msg.createReactionCollector({ filter, time: oneHour });
+    const roleReactions = await this.getFinishedReactions(msg)
+
+    collector.on('collect', async (reaction, user) => {
+      const member = msg.guild?.members.fetch(user.id)
+      if (member != undefined) {
```
prophetofxenu, line 140: I also personally find code easier to digest when there are fewer levels of nesting. Instead of `if (variable != undefined)`, I prefer `if (variable === undefined) return`. This is functionally the same, but stops you from having to create another block.

In cases like this though, `member` really shouldn't be `undefined` unless someone made a reaction and somehow left the server super quickly afterwards. So an error could also be appropriate here (as long as throwing one won't prevent future runs of the collector.)

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    collector.on('end', collected => {
+      console.log(`Collected ${collected.size} reactions.`);
+      this.sendChartToThread(msg, thread, roleReactions);
+    });
+  }
+
+  private async deleteLastChart(tc: PublicThreadChannel): Promise<void> {
+    const threadMessages = await tc.messages.fetch({ limit: 1 });
+    const latest = threadMessages.first();
```
prophetofxenu, line 170: These two lines seem like they're assuming the chart will be the only message in the thread, but I don't see anything enforcing that in the creation of the thread. Will it be enforced through permissions?

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+        // create thread and start collector
+        const thread = await this.createCTAThread(msg)
+        if (thread) {
+          this.startCTAReactionCollector(msg, thread)
+        }
+      }
+    }
+  }
+
+  private async getFinishedReactions(msg: Message): Promise<object> {
```
prophetofxenu, line 73: We should be able to specify a more specific return type here, to improve maintainability.

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+      const threadChannel = await thread;
+      if (!threadChannel.archived) {
+        await this.deleteLastChart(threadChannel);
+        await threadChannel.send({
+          files: ["./misc/tmp/reaction_summary.png"],
+          flags: [MessageFlags.SuppressNotifications]
+        });
+        this.startCTAReactionCollector(msg, thread);
+      }
+    })();
```
prophetofxenu, line 193: I think there's potential for a race condition here. The call to `writeFile` is not being awaited, so the anonymous async function below isn't going to wait for it to finish. It could finish writing before it tries to send the image, but if it doesn't, the bot may send a corrupted PNG or throw an error about another process using the file.

I think a better approach would be to use the promise version of writeFile (https://nodejs.org/api/fs.html#promise-example) which can be awaited, then run the other parts (also I'm not sure why this is wrapped inside an anonymous function, can you explain?)

apps/discord-bot/src/triggers/cta-post.ts
```diff
@@ -0,0 +1,280 @@
...
+      msg.react(finishedEmoji)
+
+      return thread
+    }
+
+    return undefined
+  }
+
+  // start a reaction collector on the given message
+  private async startCTAReactionCollector(msg: Message, thread: PublicThreadChannel): Promise<void> {
```
prophetofxenu, line 133: I see this function can be called multiple times. What happens when a collector has already been started? Does it get cleaned up? If not, could that lead to a memory leak?

apps/discord-bot/src/triggers/cta-post.ts
prophetofxenu: Even better, avoid having to write the file to disk altogether by [converting the canvas to a PNG stream](https://www.npmjs.com/package/canvas#canvascreatepngstream) and using that with an [AttachmentBuilder](https://discord.js.org/docs/packages/discord.js/main/AttachmentBuilder:Class#constructor). The builder can be passed in the array on line 188.

apps/discord-bot/.gitignore
```diff
@@ -7,6 +7,8 @@
 /config/bot-sites.json
 /config/debug.json
 
+/misc/tmp/*
```
prophetofxenu, line 10: I think we can skip having to write to the file system by using streams per my above comment, but in case we need to write to disk in the future, there are a few things I don't like about this approach:
* Relying on empty directories existing and enforcing this by committing them to the repo is a bit of a bad practice imo. Instead, use Node's functions for [checking if a directory exists](https://nodejs.org/api/fs.html#fspromisesstatpath-options), and [creating them if it does not](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options).
* Above when the path is used, it's hardcoded to use UNIX style pathing. Node also has a module for [building paths appropriate for whatever host the bot is running on](https://nodejs.org/api/path.html#pathjoinpaths).
* All OSes have a dedicated directory for temporary files, and Node has a method to get that directory on the current machine.

apps/discord-bot/package.json
```diff
@@ -48,6 +50,8 @@
     "dotenv": "^17.2.0",
     "express": "5.1.0",
     "filesize": "10.1.6",
+    "fs": "0.0.1-security",
+    "fsp": "^0.1.2",
```
prophetofxenu, line 54: * `fs` is a built in Node module and does not need to be declared as a dependency (in fact, the version here indicates this package is really a placeholder to prevent bad actors from creating a fake npm package spoofing the name.)
* fsp provides a Promise based API to the fs module, but I don't see it being used anywhere. I've suggested using Promises in other comments, but Node now has a built in Promise API with the `fs/promises` module.

apps/discord-bot/src/models/bot.ts
```diff
@@ -90,6 +95,20 @@ export class Bot {
...
     Logger.info(Logs.info.clientReady)
+
+    const ctaChannel = this.client.guilds.cache.find(dggPol => dggPol.name === guildName)?.channels.cache.find(ctaChan => ctaChan?.name === ctaChannelName);
+
+    if (ctaChannel?.type === ChannelType.GuildText) {
+      // eslint-disable-next-line no-console
+      console.log("bot started... checking for cta message...");
+      const ctaPostTrigger = new CTAPostTrigger();
+
+      ctaChannel.messages.fetch().then(msgs => {
```
prophetofxenu, line 106: Using `.then` here works, but it's already an async function, so we can also use `await` too which is a bit easier to read.

prophetofxenu (changes requested):
Logic looks solid. Most of my comments are about improving maintainability and a few potential performance concerns.

A common thing I saw was making use of `console.log`. I think using the `Logger` object we're given would be more appropriate, but I did not make a comment for every instance of this.

apps/discord-bot/src/triggers/cta-post.ts
tkmalmberg: Implemented collector tracking that checks for a collector on the current message id.

apps/discord-bot/.gitignore
tkmalmberg: Implemented the PNG stream. No longer using the file system to store the chart images.

PR #188, opened 2025-09-04
Add twitter embeds to Pragmatic Papers

prophetofxenu:
Adds a block for embedding tweets in an article.
* Supports all X oEmbed API options
* Respects light/dark theme
* Caches results using node-cache
image
image

### DIFF
--- a/apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
+++ b/apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
@@ -29,20 +29,5 @@ export const TwitterEmbed: Block = {
       ],
       defaultValue: 'none'
     },
-    {
-      name: 'lang',
-      type: 'text',
-      maxLength: 2,
-      defaultValue: 'en',
-      required: true
-    },
-    {
-      name: 'maxWidth',
-      type: 'number',
-      min: 220,
-      max: 550,
-      defaultValue: 550,
-      required: true
-    },
   ]
 }
--- a/apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
+++ b/apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
@@ -1,0 +1,48 @@
+import type { Block } from 'payload'
+
+export const TwitterEmbed: Block = {
+  slug: 'twitterEmbed',
+  interfaceName: 'TwitterEmbedBlock',
+  fields: [
+    {
+      name: 'url',
+      type: 'text',
+      required: true,
+    },
+    {
+      name: 'hideMedia',
+      type: 'checkbox',
+    },
+    {
+      name: 'hideThread',
+      type: 'checkbox',
+      
+    },
+    {
+      name: 'align',
+      type: 'radio',
+      options: [
+        { label: 'None', value: 'none' },
+        { label: 'Left', value: 'left' },
+        { label: 'Center', value: 'center' },
+        { label: 'Right', value: 'right' },
+      ],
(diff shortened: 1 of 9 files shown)


apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
```diff
@@ -0,0 +1,48 @@
...
+      required: true
+    },
+    {
+      name: 'maxWidth',
+      type: 'number',
+      min: 220,
+      max: 550,
+      defaultValue: 550,
+      required: true
+    },
```
linkion, line 46: I kind of am not liking this option to set the width of the embed, probably just leave it to us, later we can do a more proper layouting implementation

apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
```diff
@@ -0,0 +1,48 @@
...
+      name: 'align',
+      type: 'radio',
+      options: [
+        { label: 'None', value: 'none' },
+        { label: 'Left', value: 'left' },
+        { label: 'Center', value: 'center' },
+        { label: 'Right', value: 'right' },
+      ],
+      defaultValue: 'none'
+    },
```
linkion, line 31: This confused me when I first tried left align, and it was inline left aligned, with words wrapping around it

This should just be a centered block. And then for inline, there should be a separate inline block, since with the current implementation you can only inline at the beginning of a body of text, with the inline block we can inline in the middle of a body.

Perhaps you can checkout too whether we can inline word wrap in the editor with the block

apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
```diff
@@ -0,0 +1,48 @@
...
+    {
+      name: 'lang',
+      type: 'text',
+      maxLength: 2,
+      defaultValue: 'en',
+      required: true
```
linkion, line 37: Language option should be forgone, we should be using the language provided by the browser. I would be surprised if twitter doesn't already use the browser language.

linkion (dismissed):

apps/pragmatic-papers/src/blocks/TwitterEmbed/config.ts
prophetofxenu: Sorry, a little confused. Are you saying there needs to be a block and inline block version of this too?

Not sure what you mean by inline word wrap, unless you just mean this:
image

apps/pragmatic-papers/src/components/TwitterEmbed/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  lang?: string | undefined,
+  maxWidth?: number | undefined,
+}> = (props) => {
+  const [content, setContent] = useState<string>('')
+  const contentRef = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    if (!props.url) return
+
+    const theme = document.getElementsByTagName('html')[0]?.getAttribute('data-theme') ?? 'dark'
```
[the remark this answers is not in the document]
prophetofxenu: I can implement it this way, but that key isn't set when the user has Auto selected. So we'd need to assume auto means either light or dark. The current implementation is ugly but I think it's more robust.

arthurtyukayev (approved):

tkmalmberg:
@linkion do you have any changes left for this pr? If not can you clear your requested changes?

tkmalmberg (approved):

PR #197, opened 2025-09-25
Add Reddit embeds to Pragmatic Papers

prophetofxenu:
Adds a block for embedding Reddit posts in an article.

* Reddit oEmbed is not as fleshed out as X so the only input is the URL to a post.
* Caches using node-cache

image


tkmalmberg:
I'm having a problem when loading multiple (3) embeds. The first two look like they begin loading, then they seem to break before the 3rd gets onto the page. Could you look into why this might be happening?

https://github.com/user-attachments/assets/80c502f8-d009-4e4d-b0ac-f732e017483d

PR #202, opened 2025-10-13
Added census command

ZitzelAlec:
Added a census command in the same vein as how the pragmatic papers command was added.


tkmalmberg:
currently getting this when running the command:
image

tkmalmberg (changes requested):

PR #210, opened 2025-10-23
Dggpoliticalaction/setup basic homepage

ajburke:
PR to add a custom Home page to the DGG political action site

Has migrations that need to be applied before evaluating.

The Header was created with the intention of being used on other pages and its configuration settings are stored separately at http://localhost:3000/admin/globals/header by default when running locally

Likewise the configuration settings for the Home page itself are stored at http://localhost:3000/admin/globals/homepage-settings by default

The Home page will display X most recent Posts from the Posts collection, make sure you have some published to see their behavior. The Posts are colored Red, White, and Blue alternately

### DIFF
--- a/apps/dgg-political-action/src/endpoints/seed/index.ts
+++ b/apps/dgg-political-action/src/endpoints/seed/index.ts
@@ -45,9 +45,7 @@ export const seed = async ({
     globals.map((global) =>
       payload.updateGlobal({
         slug: global,
-        data: {
-          navItems: [],
-        },
+        data: {},
         depth: 0,
         context: {
           disableRevalidate: true,


apps/dgg-political-action/src/app/(frontend)/page.tsx
```diff
@@ -1,5 +1,158 @@
...
+            }}
+          >
+            {missionStatement}
+          </p>
+        </div>
+      </section>
+
+      {/* Latest News Banner */}
+      <section className="relative py-8 overflow-hidden" style={{ backgroundColor: colors.brand.red }}>              
+        {/* Left chevron */}
```
ajburke: I tried several different attempts at CSS to get these chevrons to look like chevrons but failed. I am very weak at CSS if anyone else know it well enough to take a crack at it. Leaving the html in for now until we decide if were refactoring or figuring out how to implement that part of the design

tkmalmberg: Let's not worry too much about the styling at the moment. We'll come back to it.

PR #232, opened 2025-11-30
Volunteer page function

SgainsO:
If their is a better way to make my web results see-able, let me know, but for now the database which the site details are stored is called dgg-volunteer.dbBCK


SgainsO:
Added item from media team for better design; will re-do request

PR #252, opened 2025-12-04
Revert "Add nixpacks config for pragmatic-papers to be used on Coolify"

linkion:
Reverts dggpoliticalaction/org-repo#251


linkion:
Currently drafting a concrete fix

PR #283, opened 2026-01-11
Fix slugField render behavior in admin

espdesign:
## Context

Fix for Issue:
https://github.com/dggpoliticalaction/org-repo/issues/281

Fix copied from
https://github.com/payloadcms/payload/pull/14007

Using the new module `slugField from payload` and removed old template version files `src/fields/slug`. This should help with longer term compatibility on updates.

because this uses payload official version, they renamed table columns,
```
slug_lock > generate_slug
version_slug_lock > version_generate_slug
```

When typing in the title field, the title is now properly updated and fluid, however the slug does not auto-fill anymore.

Note: I tested the behavior from an updated and new payload-website template, it seems to not auto fill on new templates as well.

## Test Plan
browse to `http://localhost:3000/admin`

Make new pages, articles, edit the new and existing titles.


tadjh:
Nice! It seems their solution was to not bother with updating the slug until you save the document.

Which makes sense and is more pragmatic than constantly updating the slug whenever the title changes, when that's not something that should change much SEO wise.

PR #287, opened 2026-01-11
bump deps

espdesign:
## Context
Needed to bump the deps to match payload version for both apps.
The dgg-political-action is behind.

As the template and site is very basic, I decided it would be worthwhile to update all deps.

## Test Plan
Don't worry about ;)


espdesign:
Did a check of versions after 3.66
https://github.com/payloadcms/payload/releases
Only noted breaking change is not implemented in our project.

linkion (approved):
LGTM, smash it

Can confirm, doesn't seem like any breaking changes were introduced in Payload, though there's a cool feature for adding statistics to admin dashboard

Issue #276, opened 2026-01-10
Pragmatic Papers Error 500 When Validating Article Body exceeded 1 MB limit


linkion:
We seem to be hitting a next js server action limit, this was found when trying to create articles on staging.

Server logs:
```
⨯ [Error: Body exceeded 1 MB limit.
To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit] {
  statusCode: 413,
  digest: '3903168031'
}
```

Client logs:

```
2460-9eb5c3033cb75435.js:1  POST https://265.staging.pragmaticpapers.com/admin/collections/articles/2 net::ERR_ABORTED 500 (Internal Server Error)
M @ 2460-9eb5c3033cb75435.js:1
await in M
A @ 2460-9eb5c3033cb75435.js:1
f @ 2460-9eb5c3033cb75435.js:1
action @ 2460-9eb5c3033cb75435.js:1
p @ 2460-9eb5c3033cb75435.js:1
(anonymous) @ 2460-9eb5c3033cb75435.js:1
dispatch @ 2460-9eb5c3033cb75435.js:1
o @ 2460-9eb5c3033cb75435.js:1
l @ 2460-9eb5c3033cb75435.js:1
(anonymous) @ 2460-9eb5c3033cb75435.js:1
t.startTransition @ 2460-9eb5c3033cb75435.js:1
(anonymous) @ 2460-9eb5c3033cb75435.js:1
l @ 2460-9eb5c3033cb75435.js:1
r @ 2460-9eb5c3033cb75435.js:1
(anonymous) @ 4eed1242-12e67c0a2cb0f582.js:1
(anonymous) @ 4eed1242-12e67c0a2cb0f582.js:1
await in (anonymous)
(anonymous) @ 4eed1242-12e67c0a2cb0f582.js:1
await in (anonymous)
(anonymous) @ 4eed1242-12e67c0a2cb0f582.js:1
T.onClick @ 4eed1242-12e67c0a2cb0f582.js:1
i8 @ e3373488-ef87eb834481fc79.js:1
(anonymous) @ e3373488-ef87eb834481fc79.js:1
nz @ e3373488-ef87eb834481fc79.js:1
sn @ e3373488-ef87eb834481fc79.js:1
cc @ e3373488-ef87eb834481fc79.js:1
ci @ e3373488-ef87eb834481fc79.js:1
[...]

PR #288, opened 2026-01-11
Add Article Resource Tracker

KyrunaKJ:
## Context

There's no current representation for authors if their article is getting too big for Payload. This adds a UI element to the sidebar when creating articles that shows how big their article is getting.
image
NOTE: This change does not adjust any validation or stop the user from inputting too much text. 

### Base Function
1. Navigate to Articles and 'Create Article'.
2. Add content to the text body.
3. If the size of the content exceeds 1MB, there will be a message displayed telling the author to cut the size of the article.
4. If they try to publish an article that is greater than 1MB, they'll hit an error.


linkion:
I see your working on #276, I think we should just go with increasing the limit and leave it at that, there shouldn't be anything user accessible for pushing data to the application, and was just showing up for writing articles, which are people we trust not to be DDOSing us.

PR #290, opened 2026-01-12
285 fix slugfield render behavior in admin

espdesign:
Duplicated https://github.com/dggpoliticalaction/org-repo/pull/283 to app/dgg-political-action

## Testing
Check imputing text on collection titles. Should be smooth now.


linkion (approved):
LGTM

PR #291, opened 2026-01-12
DGG website: update payload version

evan6seven:
## Context

I think we should keep up to date with minor payload versions


evfrenkel:
already done

Issue #189, opened 2025-09-07
User Accounts


linkion:
Setup a signup and login and profile dashboard for user accounts on the website.

PayloadCMS already has a very robust authentication and user management system, we'll just build on top of that

jkingst52:
I’d like to take this. I’m set up locally and ready can someone assign me / give me access?
or did i miss something i need to do

espdesign:
collections/users/index.ts change on admin access to : writerFieldLevel, meaning if they user cannot write in the admin panel, they should not have access to it.

tadjh:
Previous working implementation was on https://github.com/digitalgroundgame/pragmatic-papers/tree/304-3-drizzle-test but now would require major refactoring. Only useful for learning.

tadjh:
**Triage 2026-07-15:** Foundation partly advanced — a `role` field now exists on the Users collection (#710) — but no public account UI (signup / login / profile) exists yet under `src/app/(frontend)/`. Overlaps with #304 (basic-membership sign-up page); cross-linked so the two are scoped together.

PR #300, opened 2026-01-12
299 rename user role user to avoid future logic issues

### LINKED ISSUE #299: Rename User role 'user' to avoid future logic issues.
Having a User role 'user' may have potential negative effects in building future authentication logic.
Suggestion, rename to 'member' for lowest level access role.

espdesign:
## Context
Closes #299

## Test Plan

1. Confirm database has no issues with changing User role `user` to `member.`
_In theory there should be no Users yet as the feature is just now being built for them to log in._
2. Confirm new users default to 'Member' in the admin panel user creation collection.
3. Confirm user authentication in #189
_as only checkRole.ts isUser has only changed from user to member, there should be no issues with auth logic_


espdesign:
On hold, need to remove unrelated files in commit.

PR #302, opened 2026-01-12
Author feature

KyrunaKJ:
## Context

We're creating both a way for our contributors to have a place for them to provide their socials and link to other work contributed to the site, and a central page for seeing any contributors to Pragmatic Papers. Additionally, we will be displaying the authors on their associated volumes/articles for ease-of-access to the contributor's page.

#### USER TEST CASE

1) Log in to pragmatic papers admin;
2) Navigate to Authors collection;
3) Click 'Create New';
4) Add information, link User, and click Create;
    - If the User clicked is not the same User as the one creating the page, it will error and not create. They must link themselves

#### ADMIN TEST CASE

1) Log in to pragmatic papers admin;
2) Navigate to Authors collection;
3) Click 'Create New';
4) Add information, link User, and click Create;

#### FROM VOLUMES

1) Open the pragmatic papers website;
2) Click on any volume;
3) At the bottom of the page, there will be a section called 'Meet the Author(s)' that will have a card for each Author of any article in the volume. Their page can be navigated to from here.
image

#### FROM AUTHORS

1) Open pragmaticpapers.com/authors;
2) Any Author created on the site will be listed here as a card. Their page can be navigated to from here.
image

#### FROM ARTICLE

1) Open any pragmatic papers article, either from a slug or volume.
2) At the bottom of the article, if there was an Author created for a contributor, there will be a Meet the Author section. Their page can be navigated to from there. If an Author was not created, then there will be no section.
image

### AUTHOR PAGE

1) Navigate to any author page;
2) If provided on creation, the page will contain the author's profile picture, socials, bio, and a list of any associated article.
image

### DIFF
--- a/apps/pragmatic-papers/src/migrations/20260304_215356_feature_authors.json
+++ b/apps/pragmatic-papers/src/migrations/20260304_215356_feature_authors.json
@@ -1,0 +1,11040 @@
+{
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.pages_hero_links": {
+      "name": "pages_hero_links",
+      "schema": "",
+      "columns": {
+        "_order": {
+          "name": "_order",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "_parent_id": {
+          "name": "_parent_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "id": {
+          "name": "id",
+          "type": "varchar",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "link_type": {
+          "name": "link_type",
+          "type": "enum_pages_hero_links_link_type",
+          "typeSchema": "public",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "'reference'"
+        },
+        "link_new_tab": {
+          "name": "link_new_tab",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "link_url": {
+          "name": "link_url",
+          "type": "varchar",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "link_label": {
+          "name": "link_label",
+          "type": "varchar",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "link_appearance": {
+          "name": "link_appearance",
+          "type": "enum_pages_hero_links_link_appearance",
+          "typeSchema": "public",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "'default'"
+        }
+      },
+      "indexes": {
+        "pages_hero_links_order_idx": {
+          "name": "pages_hero_links_order_idx",
+          "columns": [
+            {
+              "expression": "_order",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "pages_hero_links_parent_id_idx": {
+          "name": "pages_hero_links_parent_id_idx",
+          "columns": [
+            {
+              "expression": "_parent_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "pages_hero_links_parent_id_fk": {
+          "name": "pages_hero_links_parent_id_fk",
+          "tableFrom": "pages_hero_links",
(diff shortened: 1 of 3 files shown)


apps/pragmatic-papers/src/app/(frontend)/authors/[slug]/page.tsx
Comment (tadjh) [inline apps/pragmatic-papers/src/app/(frontend)/authors/[slug]/page.tsx]: add aria-labels to Link tags

apps/pragmatic-papers/src/app/(frontend)/articles/[slug]/page.tsx
```diff
@@ -88,6 +107,31 @@ export default async function Article({ params: paramsPromise }: Args): Promise<
       <ArticleHero article={article} />
 
       <RichText className="" data={article.content} enableGutter={false} />
+
```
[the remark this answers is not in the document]
Comment (tadjh) [reply apps/pragmatic-papers/src/app/(frontend)/articles/[slug]/page.tsx]: Here's design inspiration for what I'm thinking [Author Card](https://v0.app/chat/author-block-hp0x52pllvB?ref=HHYQTK) feel free to copy it and mold it to you liking

apps/pragmatic-papers/src/app/(frontend)/articles/[slug]/page.tsx
```diff
@@ -72,7 +97,12 @@ export default async function Article({ params: paramsPromise }: Args): Promise<
 
   if (!article) return <PayloadRedirects url={url} />
 
-  const { footnotes, content } = article
+  const populatedAuthors = article.populatedAuthors || []
+  const authorIds = populatedAuthors
+    .map((author) => author?.id)
+    .filter((id): id is string => typeof id === 'string')
+
+  const authors = await queryAuthorsByIds({ ids: authorIds })
```
Comment (tadjh) [inline apps/pragmatic-papers/src/app/(frontend)/articles/[slug]/page.tsx]: it looks like the article already populates the authors on `article.authors`

apps/pragmatic-papers/src/components/Articles/AuthorArticleCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <Card className={cn('h-full', className)}>
+      <CardContent ref={card.ref} className="flex flex-row gap-4 p-6 sm:flex-row">
+        <div className="h-24 w-32 flex-shrink-0 overflow-hidden rounded border border-border bg-muted sm:h-28 sm:w-40">
+          {metaImage && typeof metaImage !== 'string' && (
+            <Media
+              resource={metaImage}
+              className="h-full w-full"
+              imgClassName="h-full w-full object-cover"
+              size="square"
```
tadjh: Remove `size="square"` as it breaks the image layout

apps/pragmatic-papers/src/collections/Users/index.ts
```diff
@@ -81,6 +146,88 @@ export const Users: CollectionConfig = {
...
+    {
+      name: 'loginAttempts',
+      type: 'number',
+      access: {
+        read: () => false,
+      },
+      admin: {
+        disabled: true,
+      },
+    },
```
tadjh: Why are these added?

apps/pragmatic-papers/src/collections/Users/index.ts
```diff
@@ -81,6 +146,88 @@ export const Users: CollectionConfig = {
...
   ],
   timestamps: true,
+  hooks: {
+    beforeChange: [
+      (args: Parameters<CollectionBeforeChangeHook<User>>[0]): Partial<User> | void => {
```
tadjh: Fix typing use `CollectionBeforeChangeHook<User>` to simplify all of the typing. See `generateFootnotes`

apps/pragmatic-papers/src/collections/Articles/index.ts
tadjh: Why was this changed? What are the implications for returning a Date type versus an iso string?

apps/pragmatic-papers/src/app/(frontend)/authors/[slug]/page.tsx
```diff
@@ -0,0 +1,310 @@
...
+                  href={normalized}
+                  target="_blank"
+                  rel="noreferrer noopener"
+                  className="text-sm text-brand underline-offset-2 hover:underline"
+                  aria-label={label ? `${label} (opens in a new tab)` : 'External author link'}
+                >
+                  {label}
+                </a>
+              )
+            })}
```
Comment (tadjh) [inline apps/pragmatic-papers/src/app/(frontend)/authors/[slug]/page.tsx]: Probably should show same icons we see on AuthorCard, so reuse the icon map that I recommend below.

tadjh (approved):

apps/pragmatic-papers/src/collections/Users/index.ts
tadjh: When i access `http://localhost:8000/authors/writer-two` while Logged out I am able to see the author information while removing these fields. When I console.log out the User object, these fields are not present. I don't think we need these.

Issue #32, opened 2025-07-02
Richtext feature: Support for citations with bibliography on bottom of article or footnotes


linkion:
This can be implement later, but would be nice to have

Basically an inline block for linking to citations or footnotes[^1]

[^1]: Here's the inline footnote definition.

PR #309, opened 2026-01-15
PP feature: footnotes

evan6seven:
#32 

Implemented inline footnotes end‑to‑end: added the Footnote block, wired it into the Articles Lexical editor, and updated rich text rendering to collect inline footnotes, number them, render a bottom list, and provide forward/back links. Notes shown on hover also.

New block + admin label: config.ts, AdminComponent.tsx
Frontend rendering + numbering + list + backlinks: index.tsx, Component.tsx
Editor wiring: index.ts

https://github.com/user-attachments/assets/94c5cb8f-02c9-4728-8045-d1195ff6b102


tadjh:
Looks good at first glance gonna pull this and take a look in the morning!

apps/pragmatic-papers/src/blocks/Footnote/Component.tsx
```diff
@@ -0,0 +1,39 @@
...
+  blockType: 'footnote'
+}
+
+type FootnoteInlineProps = FootnoteBlockProps & {
+  className?: string
+  index?: number
+}
+
+export const FootnoteBlock: React.FC<FootnoteInlineProps> = ({ note, index, className }) => {
+  if (!note || !index) return null
```
tadjh: Index can be 0 which is falsy. Check for not undefined explicitly.

apps/pragmatic-papers/src/blocks/Footnote/Component.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <sup
+      className={cn('footnote', className)}
+      title={note}
+      aria-label={`Footnote: ${note}`}
+      id={referenceId}
+    >
+      {targetId ? (
+        <a className="footnote-link text-brand no-underline shadow-none" href={`#${targetId}`}>
```
tadjh: is footnote-link unused? Not seeing it's styles on mobile if so

tadjh (dismissed):
Think footnotes should come out of the database sorted. Would remove the need for the tree walker, currying and index map in one fell swoop. Just render index + 1 directly in the tsx since they'd be sorted.

evan6seven:
As far as I understand, inline blocks are only stored in the database inside of the JSON tree of the rich text. Therefore, we must walk the tree in order to compute the index order / build a footnote array on the article. 

Everything else you mentioned, I've updated.

tadjh:
Hey @evan6seven made some changes for you to look over. 

- Add database migrations
- Add footnotes to Articles RSS Feed
- Add attribution links
- Add smooth scroll
- Removed backlinks as footnotes should instead show attribution links (they were nice to have just was creating confusion with so many links in the footer of each article, users wont know where to click and get frustrated when instead the arrow takes them up the page instead of the source attribution) 
- Improved reusability of code by creating footnotesFields so that the Rich Text representation and the generated footnotes are the same shape.
- Cleaned up some of the separation of concerns (using a FootnoteList instead of inserting the footnotes in the article manually.)

Seems to work great! Can you check over my work and see if I broke anything?

evan6seven:
Seems good.

One thing, should the title of the attribution be whats shown on the page instead of the url? I think the title is primarily what someone wants to see rather than the url.

evan6seven:
In that case I think the title attribute should be the same as the note text (automatically), or there shouldn't be a a title attribute.

tadjh:
Fixed those @evan6seven

tadjh (approved):
Think this is ready to go!

tadjh:
> @tadjh I noted something while testing this. I have one footnote but the aria-label was "Link to source 4". It should match the footnote number right?

Oo good catch! Yes its should, could you commit the fix? The index attribute should be passed to that component that displays the link props.

> Also the footnote was positioned out to the left of its area still. 

Oh man thought I pushed it over enough. Search for `[18px]` and you should be able to bump it up till it's aligned.

Issue #310, opened 2026-01-15
Move Google Analytics ID in env vars


tadjh:
- Add `NEXT_PUBLIC_GOOGLE_ANALYTICS_ID` or something like that
- Replace hardcoded value on line 84 in `apps\pragmatic-papers\src\app\(frontend)\layout.tsx`
- Update .env.example
- Update README if necessary
- Add `NEXT_PUBLIC_GOOGLE_ANALYTICS_ID` to environments.d.ts
- Maybe need to add to turbo.json in root as well?

PR #314, opened 2026-01-16
PP: google id to env var

evan6seven:
#310 

shrimple as

requires env update in prod

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,10 @@ pnpm dev
 ```
 5. Use a browser to view the applications,
 
-```
-http://localhost:8000
-```
+  - pragmatic papers : `http://localhost:8000`
+  
+  - dgg-political-action:`http://localhost:8888`
 
-```
-http://localhost:8000
-```
 
 > [!TIP] 
 > You can also start a single application by running `pnpm dev` in the applications root directory.


apps/pragmatic-papers/src/environment.d.ts
```diff
@@ -3,6 +3,7 @@ declare global {
     interface ProcessEnv {
       PAYLOAD_SECRET: string
       DATABASE_URI: string
+      NEXT_PUBLIC_GOOGLE_ANALYTICS_ID?: string
```
tadjh: remove the ? on this type. Its safe to say this env var exists, otherwise the build would have already failed.

apps/pragmatic-papers/src/app/(frontend)/layout.tsx
```diff
@@ -74,7 +75,7 @@ export default async function RootLayout({ children }: { children: React.ReactNo
           <Header />
           {children}
           <Footer />
-          <GoogleAnalytics gaId="G-PXK2QL92HV" />
+          {googleAnalyticsId ? <GoogleAnalytics gaId={googleAnalyticsId} /> : null}
```
tadjh: Once you change environments.d.ts this ternary wont be required as googleAnalyticsId will always exist.

PR #316, opened 2026-01-16
feat(pragmatic-papers): Add Iceout redirect

KyrunaKJ:
## Context

We are creating stickers that will direct users to Iceout's report page. This code sets up a redirect from our site so we can have multiple QR codes generated with different state suffixes and not have to remake all these stickers in case their site changes their structure.

## Test Plan

1) Navigate to https://www.pragmaticpapers.com/iceout/mn
1a) You will be redirected to https://iceout.org/en/location/report

### DIFF
--- a/apps/pragmatic-papers/redirects.js
+++ b/apps/pragmatic-papers/redirects.js
@@ -13,7 +13,7 @@ const redirects = async () => {
   }
 
   const iceoutRedirect = {
-    source: '/iceout/:state',
+    source: '/iceout/:state*',
     destination: 'https://iceout.org/en/location/report',
     permanent: false,
   }


apps/pragmatic-papers/redirects.js
```diff
@@ -12,7 +12,13 @@ const redirects = async () => {
     source: '/:path((?!ie-incompatible.html$).*)', // all pages except the incompatibility page
   }
 
-  const redirects = [internetExplorerRedirect]
+  const iceoutRedirect = {
+    source: '/iceout/:state',
```
tadjh: going to `https://localhost:8000/iceout` leads to a 404.

this source means the `:state` param is not optional.

change it to `:state*` to handle the case where someone doesn't type a state.

tkmalmberg (approved):
This looks good to me. Please make the changes @tadjh is looking for to merge.

tadjh (approved):

PR #335, opened 2026-01-25
Command role checks

Flargebla:
## Context

This PR adds a generic role checking mechanism for current and future commands. Commands can be restricted to specific server roles now. This is a run-time enforcement so out-of-role users will be able to begin executing the command but it will return a permission denied.

This PR also adds the Coordinator role requirement for running "Send Dev Onboarding" command.

## Test Plan

To test this, a Coordinator should try running the Send Dev Onboarding command. Then, a non-Coordinator should try running the command. It should work for the first person, and the other should be rejected.

### DIFF
--- a/apps/discord-bot/config/config.json
+++ b/apps/discord-bot/config/config.json
@@ -75,5 +75,27 @@
     "rateLimit": {
       "minTimeout": 30
     }
+  },
+  "roles": {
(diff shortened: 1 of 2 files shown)


PaulSearcy (approved):
LGTM

PR #349, opened 2026-01-29
Merge branch 'main' of https://github.com/dggpoliticalaction/org-repo…

tkmalmberg:
… into dev

## Context

Updates dev to align with main.


apps/dgg-political-action/eslint.config.mjs
```diff
@@ -1,4 +1,9 @@
 import { nextJsConfig } from '@repo/eslint-config/next-js'
 
 /** @type {import("eslint").Linter.Config} */
-export default nextJsConfig
```
evan6seven: Left side is correct here

PR #352, opened 2026-02-10
Fix/react bugs

KyrunaKJ:
## Context

Was wanting to go through the pragmatic papers project and see if there were any common React issues the site could have.

## Test Plan

Each commit in this PR can be tested without the change on a fresh dev branch and then be cherry-picked for testing (or they can be tested all at once).

For Replace Index as Key: prior to changes, affected components will have their keys be their index. This can cause render errors if items are added or removed from the list they're created in. React prefers having keys that won't update so we are changing to use ids where we can or some created key derived from the object's fields.

For tighten object dependencies: prior to changes, any change made to an object will cause the whole object to re-render. In changed objects, we won't rerender every field in an object when one field is changed.

For add stricter react hooks linting: this is just a forward-looking change, and should help development as more devs onboard and use AI for development. This will keep our rendering safe by throwing linting errors if unsafe code is written.

### DIFF
--- a/apps/dgg-political-action/src/components/Card/index.tsx
+++ b/apps/dgg-political-action/src/components/Card/index.tsx
@@ -52,7 +52,7 @@ export const Card: React.FC<{
 
                     const categoryTitle = titleFromCategory || 'Untitled category'
 
-                    const key = (category as { id?: string | number }).id ??
+                    const key = (category as Category).id ??
                       categoryTitle ??
                       index.toString()
 


apps/dgg-political-action/src/components/Card/index.tsx
```diff
@@ -52,10 +52,14 @@ export const Card: React.FC<{
 
                     const categoryTitle = titleFromCategory || 'Untitled category'
 
+                    const key = (category as { id?: string | number }).id ??
```
deckmasterbeam: This partial typing gets the job done in terms of letting TS know that this object is expected to have an id. But, this partial typing can be confusing to work with. This partial type is consistent with the definition of the type Category, what if you just used that type here?
```suggestion
                    const key = (category as Category).id ??
```
A couple of places follow this same partial typing pattern. Could also be spots to use a declared type

apps/dgg-political-action/src/components/CollectionArchive/index.tsx
```diff
@@ -18,8 +18,11 @@ export const CollectionArchive: React.FC<Props> = (props) => {
         <div className="grid grid-cols-4 sm:grid-cols-8 lg:grid-cols-12 gap-y-4 gap-x-4 lg:gap-y-8 lg:gap-x-8 xl:gap-x-8">
           {posts?.map((result, index) => {
             if (typeof result === 'object' && result !== null) {
```
deckmasterbeam: After a check like this, you could consider re-casting posts as something like `const postsObj = posts as Posts[];`, lets you clear up some of the typing below

apps/pragmatic-papers/src/Header/Nav/index.tsx
```diff
@@ -11,8 +11,9 @@ export const HeaderNav: React.FC<{ data: HeaderType }> = ({ data }) => {
 
   return (
     <nav className="flex items-center gap-3">
-      {navItems.map(({ link }, i) => {
-        return <CMSLink key={i} {...link} appearance="link" />
+      {navItems.map(({ id, link }) => {
```
deckmasterbeam: Think this is a good candidate to memoize? The map may need to get recomputed if/when data changes?

apps/dgg-political-action/src/blocks/Contributors/Component.tsx
```diff
@@ -53,12 +54,13 @@ export const ContributorsBlock: React.FC<ContributorsBlockProps> = ({ title, con
...
                   {socialLinks.map((link: SocialLink, linkIndex: number) => {
-                    const { platform, url } = link
+                    const { platform, url, id } = link
+                    const linkKey = id ?? `${platform}-${url || linkIndex}`
                     const Icon = platformIcons[platform as keyof typeof platformIcons]
 
                     return (
                       <a
-                        key={linkIndex}
+                        key={linkKey}
```
tadjh: `link.id || lineIndex` should suffice as link.id always exists.

apps/dgg-political-action/src/blocks/RelatedPosts/Component.tsx
```diff
@@ -29,7 +29,9 @@ export const RelatedPosts: React.FC<RelatedPostsProps> = (props) => {
         {docs?.map((doc, index) => {
           if (typeof doc === 'string') return null
 
-          return <Card key={index} doc={doc} relationTo="posts" showCategories />
+          const key = doc.id ?? doc.slug ?? index.toString()
+
+          return <Card key={key} doc={doc} relationTo="posts" showCategories />
```
tadjh: `doc.id || index` as doc.id always exists.

apps/dgg-political-action/src/Footer/Component.tsx
```diff
@@ -24,8 +24,9 @@ export async function Footer() {
         <div className="flex flex-col-reverse items-start md:flex-row gap-4 md:items-center">
           <ThemeSelector />
           <nav className="flex flex-col md:flex-row gap-4">
-            {navItems.map(({ link }, i) => {
-              return <CMSLink className="text-white" key={i} {...link} />
+            {navItems.map(({ id, link }) => {
+              const key = id ?? link?.url ?? link?.label
+              return <CMSLink className="text-white" key={key} {...link} />
```
tadjh: `link.id || index`

apps/dgg-political-action/src/Header/Nav/index.tsx
```diff
@@ -13,8 +13,9 @@ export const HeaderNav: React.FC<{ data: HeaderType }> = ({ data }) => {
 
   return (
     <nav className="flex gap-3 items-center">
-      {navItems.map(({ link }, i) => {
-        return <CMSLink key={i} {...link} appearance="link" />
+      {navItems.map(({ id, link }) => {
+        const key = id ?? link?.url ?? link?.label
+        return <CMSLink key={key} {...link} appearance="link" />
```
tadjh: `link.id || index`

tadjh (commented):
Use id's that are already provided from the db if we want to update to more stable keys appended with array index. Incomplete typing are hiding the fact that the actual types in payload-types.ts include stable ids for everything we want to render. The template files that came with the payload repo didn't depend on payload-types.ts for whatever reason so they've persisted until now. By way of updating these keys we should be using the generated payload types.

PR #385, opened 2026-03-07
Refactor away from custom theme provider

TedTimbrell:
## Context

Part of https://github.com/dggpoliticalaction/org-repo/issues/363 where we want to move away from the custom theme provider

## Test Plan

Nothing like walking through it all yourself


TedTimbrell:
closes #392

PR #388, opened 2026-03-08
Swap out logos, favicons, opengraph image

espdesign:
## Context
https://github.com/dggpoliticalaction/org-repo/issues/380

## Test Plan

- [x]     Confirm "Change tailwind color text-brand to #ff401a" NOTE: there was two different blue colors to change, adjusted a similar orange to fit.
- [x]     Confirm text brand is set correctly - "Replace style="fill: #ff401a" with classname="text-brand""
- [x]     Confirm no existing DggLIcon - "Deprecate LoveIcon.tsx"
- [x]     Confirm resizing the logo looks right - "Update the tailwind height classes for best responsive sizes."


apps/pragmatic-papers/public/pragmaticpapers-logo-light.svg
tadjh: I think we delete these files completely as I believe they are unused now.

PR #396, opened 2026-03-09
Basic Recommendation Algorithm

### LINKED ISSUE #308: Add Related Articles
Increase site engagement by showing related articles at the bottom of each article.

Use Volume sibilings, Article Titles and Topics to surface relevant results.

Ideally this would be static, meaning related articles are generated with a click of a button (either a RelatedArticleBlock or on save) and save with each article.

Do it the payload way: The article collection might get a new `relatedArticles` relationship field with hasMany set to true, etc. Use a collection hook to update on save?

TedTimbrell:
## Context

First attempt at a recommendation algorithm. Articles with a high read rate (users that scrolled 90% of the page over total users) with a recency decay.

It filters out articles with less than 10 views except if it's the most recent volume. 

## Test Plan
This is pretty awkward to test as it requires relevant GA + production data. I'll remove the UI components from this PR so I can run some generation tests when merged into prod.


scripts/updateRecommendations.ts
```diff
@@ -0,0 +1,66 @@
+/*
+MUST READ: This is a debug script for playing with recommendation algorithm logic. 
+See src/jobs/updateReccomendations.ts for the ACTUAL CRON.
```
TedTimbrell:
```suggestion
See src/jobs/updateRecommendations.ts for the ACTUAL CRON.
```

tadjh (approved):
LGTM!

PR #397, opened 2026-03-09
Add Claude.md

TedTimbrell:
## Context
The PR description is actually written by a human

This file is largely generated by claude code itself based on its investigation of the codebase with some manual edits by me. A couple of important notes:
- I think it's okay for this sort of file to be out of date. It's pretty trivial for someone using claude code to have it check the codebase for updates and make changes
- It is weird that there is so much coupling of comment and logic but the goal of the file is to give an agent succinct context. For example:
    - claude code doesn't know that we have such strict rules on lint warnings. With this file and note, it (should) run linting itself before completing a change.
    - We explicitly note that new ignores are mostly not allowed. Theoretically, this will help new contributors from having to go back and edit because their agent was being super lazy.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,14 +26,12 @@ Turborepo monorepo for **Pragmatic Papers**, a Next.js 15 website with Payload C
   - `pnpm payload generate:types` — regenerate Payload TypeScript types
   - `pnpm payload generate:importmap` — regenerate Payload import map
   - `pnpm payload migrate` — run database migrations
-  - `pnpm payload migrate:create` — create a new migration
-
+  - `pnpm payload migrate:create "migation_name"` — create a new migration. Pass a name as the first argument.
 
 ## Architecture
 
 ### Monorepo Structure
 - **`apps/pragmatic-papers/`** — Main Next.js + Payload CMS application
-- **`packages/ui/`** — Shared React component library (button, card, code)
 - **`packages/eslint-config/`** — Shared ESLint configs (base, next-js, react-internal)
 - **`packages/typescript-config/`** — Shared tsconfig presets (base, nextjs, react-library)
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,137 @@
...
+- **Styling**: TailwindCSS with CSS variables (HSL) for theming
+- **Content rendering**: Blocks system with Lexical rich text editor; each block has a config and a React component
+- **Pre-commit hooks**: Husky + lint-staged runs Prettier and ESLint (`--max-warnings 0`) on staged files
+- **Colocation**: Prefer colocating logic near where it's used. `src/utilities/` is only for genuinely reusable helpers shared across multiple features (e.g. `generateMeta`, `getURL`, `toRoman`, `cn`). Don't put single-use logic there.
+
+### Requirements
+- Node.js >= 22
+- pnpm 10
+- Docker (for local PostgreSQL on port 9000)
+- No test framework set up yet
```
TedTimbrell:
```suggestion
### Testing your changes
- run linting and type-checks
- No unit or integration test framework set up yet - ignore for now
```

AGENTS.md
```diff
@@ -0,0 +1,137 @@
...
+Key env vars are defined in `apps/pragmatic-papers/.env.example`:
+- `DATABASE_URI` — Postgres connection string
+- `PAYLOAD_SECRET` — JWT encryption
+- `NEXT_PUBLIC_SERVER_URL` — CORS, links (no trailing slash)
+- `NEXT_PUBLIC_GOOGLE_ANALYTICS_ID` — Google Analytics
+- `CRON_SECRET` — cron job auth
+- `PREVIEW_SECRET` — preview request validation
+- `USE_LOCAL_STORAGE` — `true` for local file storage, otherwise S3
+- S3 (production): `S3_ENDPOINT`, `S3_REGION`, `S3_BUCKET`, `S3_ACCESS_KEY_ID`, `S3_SECRET_ACCESS_KEY`
+- `NEXT_PUBLIC_SUPABASE_URL` — used for public media URLs when using S3 via Supabase
```
TedTimbrell:
```suggestion
```

I think this is too much and kinda wrong. If we want to comment on the env setup in turbo repo and duplicating to .env.example for dev work we should

Issue #433, opened 2026-03-11
Add Pagination to Author Single Page


tadjh:
Some Authors have hundreds of articles on prod.

We should add pagination to the author single page at `apps\pragmatic-papers\src\app\(frontend)\authors\[slug]\page.tsx`.

Use searchparams like `?p=2` to paginate the the database query `queryArticlesByAuthor`.

References:
https://payloadcms.com/docs/queries/pagination 
https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional
https://nextjs.org/docs/app/api-reference/file-conventions/page#page-props-helper

Note: when we upgrade shadcn Pagination component we will move to using `href` instead of `onClick` to navigate. We will replace `a` with `Link` from `next/link` as well and consider using `HoverPrefetchLink` under the hood to keep next from prefetching every page while retaining smooth client side routing.

PR #438, opened 2026-03-12
Author Article Pagination

KyrunaKJ:
## Context

#433 

## Test Plan
image

p was added as a search param. We resolve p to 1 if an invalid/non-Number input is given. Otherwise we show 5 results per page.

Author Article Cards were also updated to have a 2/2/1 line clamp format; the title was previously unconstrained and would crowd the other elements if it was too long.


tadjh:
@kyruna sorry about the merge conflicts, I can fix them if you need me to!

PR #441, opened 2026-03-12
Delete `useClickableCard`

TedTimbrell:
## Context

Css provides the tools to spread the link for us. Using a client component here means the link doesn't work until the JS loads. 

The css approach is a bit tricky to work with but is just a better approach given almost everything we do is SSR'd.


tadjh:
Good catch! Did I merge those in? Those shouldn't have made it through.

PR #452, opened 2026-03-14
Optimize ImageMedia and Refactor Lightbox

tadjh:
## Context

Images are slow and buggy. Now they are much smoother.

## Test Plan

Seed and check all pages with images.


AlexHagood (dismissed):
thanks!

AlexHagood (approved):

Issue #512, opened 2026-03-23
Fix og:url for Articles


mca10:
All articles I checked have `og:url` set to the root url, example:

<img width="754" height="644" alt="Image" src="https://github.com/user-attachments/assets/b235e359-9d74-44ec-8315-9d7a6bce1b7a" />

The og:url metadata needs to point to the correct canonical url

PR #515, opened 2026-03-23
448+512 Add SEO/opengrpah data to topics pages and fix og:url tags

### LINKED ISSUE #488: SEO: Topics pages do not have an opengraph image
Copy over SEO Plugin config from Article Collection to Topics.

Example:
Image
See other Examples in code.

Image

mca10:
Fixes #488 and #512

### DIFF
--- a/apps/pragmatic-papers/src/utilities/generateMeta.ts
+++ b/apps/pragmatic-papers/src/utilities/generateMeta.ts
@@ -28,7 +28,11 @@ export const generateMeta = async (args: {
           ]
         : undefined,
       title,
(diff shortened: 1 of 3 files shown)


apps/pragmatic-papers/src/utilities/generateMeta.ts
```diff
@@ -26,7 +28,7 @@ export const generateMeta = async (args: {
           ]
         : undefined,
       title,
-      url: Array.isArray(doc?.slug) ? doc?.slug.join("/") : "/",
```
mca10: Bit of a judgement call here, I don't really understand why were we doing `slug.join` seeing that slug is a string ?

Instead `generateMeta` receives an explicit canonical relative URL that gets appended to the server URL, with a fallback for whatever reason `doc.slug` was being called.

Sanity check welcomed!

PR #550, opened 2026-03-26
Restore mergeOpenGraph in Root Layout

tadjh:
## Context
 
I removed it to see if Production behaved any differently, since the double SEO bug only appears in Production.

Removing it had no noticeable effect.

Starting to think it's cloudflare.


tadjh:
Still learning.

PR #556, opened 2026-03-27
Hotfix: Restore Opengraph (#553)

tadjh:


tadjh:
I think the last one didn't work actually since main is currently 1 commit ahead and 1 commit behind. I consulted our good buddy Claude and apparently one of the steps was wrong. I reused the hotfix branch direct to dev, instead of creating a new branch that merges origin/main into dev.

The (#553) (#554) is actually a bad sign. @espdesign

PR #575, opened 2026-04-05
Replace next/link with plain <a> tags for Cloudflare caching

TedTimbrell:
## Summary

- Replace all `next/link` `<Link>` components with plain `<a>` tags across the frontend
- Disable the `@next/next/no-html-link-for-pages` ESLint rule since we're intentionally not using `next/link`

## Why

While cloudflare seems to be caching our pages, the RSC query param is rarely the same when going from one page to the next, which makes most internal navigation cold/MISS.

Since our origin server is slow, uncached client-side navigations take several seconds. By switching to plain `<a>` tags, all navigations become full page loads that Cloudflare can cache at the edge, making page-to-page navigation significantly faster for users.

## Tradeoff

- No more client-side (SPA-style) navigation — every link click is a full page load
- This is a net win because cached full page loads from Cloudflare's edge are much faster than uncached RSC requests hitting our origin

## Test plan

- [ ] Verify all nav links in header/footer work correctly
- [ ] Verify article links, volume links, and topic links navigate properly
- [ ] Verify footnote anchor links still scroll to the correct position
- [ ] Check Cloudflare dashboard — page responses should now show `cf-cache-status: HIT` on repeat visits
- [ ] Confirm `Vary` response header no longer includes RSC-related values on page responses


apps/pragmatic-papers/src/components/Link/HoverPrefetchLink.tsx
tadjh: Lets leave this file as is, and simply not use HoverPrefetchLink within the app.

PR #601, opened 2026-04-23
Add breadcrumbs

### LINKED ISSUE #493: Breadcrumbs
Starting to get claustrophobic when I'm navigating deeper into the site.

Just need some bread... crumbs to make me feel better.

- [x] use shadcn `breadcrumb` component. https://ui.shadcn.com/docs/components/base/breadcrumb. We use `base-ui` not `radix`.
- [ ] Likely need to create a `layout.tsx` in each route you want to add breadcrumbs to this way you have access to the path through searchparams. Or `useSelectedLayoutSegment` to make a general use one?

gsabado:
## Context

Addresses issue #493 

Adds generic breadcrumb block to root page


src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+// constants
+const SECTION_DROPDOWN_LINKS = [
+  { href: "/articles", label: "Articles" },
+  { href: "/topics", label: "Topics" },
+  { href: "/volumes", label: "Volumes" },
+  { href: "/authors", label: "Authors" },
+] as const
+const CURRENT_PAGE_CLASS = "block min-w-0 max-w-full truncate"
+const DROPDOWN_TRIGGER_CLASS = "flex items-center gap-1 capitalize"
+const DROPDOWN_CONTENT_CLASS = "w-max min-w-36"
```
tadjh: Lets move these classes into the component attributes. This way they properly lint, format, and provide intellisense.

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+                      <ChevronDown className="h-4 w-4" />
+                    </DropdownMenuTrigger>
+                    <DropdownMenuContent className={DROPDOWN_CONTENT_CLASS}>
+                      {dropdownLinks.map((link) => (
+                        <DropdownMenuItem key={link.href}>
+                          <a href={link.href}>{link.label}</a>
+                        </DropdownMenuItem>
+                      ))}
+                    </DropdownMenuContent>
+                  </DropdownMenu>
```
tadjh: I'm thinking no additional dropdown list for now. This is nice, but I don't think it's something we need at the moment.

tadjh (commented):
Looking good.

Upon browsing around, I think I like the breadcrumbs where we have them except for on  individual `articles` for now. Need to think on the desired presentation for Articles in particular.

src/components/Breadcrumbs/index.tsx
gsabado: inlined!

src/components/Breadcrumbs/index.tsx
gsabado: removed

gsabado:
removed article name breadcrumb, the other depth 3 breadcrumbs still appear, for example "Topic > Nature" and "Volumes > 4"

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    href: getSegmentHref(segments, index),
+    label: formatBreadcrumbLabel(segment),
+  }))
+
+export const Breadcrumbs = (): ReactElement | null => {
+  let segments = useSelectedLayoutSegments()
+
+  // if there are no segments, we're on the homepage, so we don't need to render breadcrumbs
+  // if the first segment is "articles", we want to show only "Articles" in the breadcrumbs, so we slice the segments array to keep only the first segment
+  if (segments.length === 0) return null
```
tadjh:
```suggestion
  if (segments.length === 0 || segments[0] === "articles") return null
```

tadjh:
Extract queries to utilities and add middleware
Move repeated payload query helpers (queryUserBySlug, queryTopicBySlug, queryVolumeBySlug, queryPageBySlug) into a new utilities/queries.ts and replace local cached implementations with imports. Add a middleware that injects an x-pathname header so server-side Breadcrumbs can resolve human-friendly labels; Breadcrumbs was refactored to be a server component, uses headers + Suspense with a skeleton fallback, and looks up names via the centralized queries. Minor UI/layout tweaks: update container widths in VolumeViewBlock and PageHero, and reorder/export identifiers in the breadcrumb UI module. This centralizes data fetching logic, reduces duplication, and enables server-resolved breadcrumb labels.

Client side breadcrumbs with useLayoutSegments proved to be unsatisfactory.

x-pathname is on request, and can't be spoofed. Matcher used to avoid sending this on requests this header is not needed for.

tadjh (approved):
LGTM

tadjh (approved):

PR #610, opened 2026-04-27
Footnote Label Update

TedTimbrell:
## Context

The footnote UI is a little clunky
* Footnote numbers were only generated on a full save which only Editor-roled users were able to do
* Even still, the payload update was sent down to the client before the number was generated, so instead we opt for the first 20 characters of the citation. 
* Adds a `[^<text>]` pattern to auto convert to a footnote

### DIFF
--- a/src/blocks/Footnote/FootnoteLabel.tsx
+++ b/src/blocks/Footnote/FootnoteLabel.tsx
@@ -5,10 +5,16 @@ interface FootnoteLabelProps {
   siblingData: Partial<FootnoteBlock>
 }
 
+const PREVIEW_LIMIT = 20
+
+const truncate = (text: string): string =>
+  text.length > PREVIEW_LIMIT ? `${text.slice(0, PREVIEW_LIMIT)}…` : text
+
 export const FootnoteLabel: React.FC<FootnoteLabelProps> = ({
   siblingData: { index, sourceId, note },
 }) => {
   if (!note) return <span>Footnote</span>
+  if (typeof index !== "number") return <span>{truncate(note)}</span>
   if (sourceId) return <span>{`↗ [${index}]`}</span>
   return <span>{`[${index}]`}</span>
 }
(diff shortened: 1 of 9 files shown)


src/collections/Articles/hooks/generateFootnotes.ts
tadjh, line 1: Thoughts on collecting footnotes on every autosave? Worried about performance on a full article with many sources.

src/blocks/Footnote/FootnoteLabel.tsx
tadjh, line 1: Can we keep numbers and instead generate the index on the field level beforeChange hook, instead of on the collection hooks. Jacob mentioned missing the numbers and the label text is much harder to reason about than the way they look on the front end.

TedTimbrell: Reading around what Jacob said, I think it's mostly that it just said "footnote" which made it indistinguishable and that the only other option was it being the number. Unless you're referencing something in the past?

src/collections/Articles/hooks/generateFootnotes.ts
TedTimbrell: With the 800 ms debounce I feel like it'll be fine?

src/blocks/Footnote/FootnoteLabel.tsx
tadjh: We currently show numbers not the contents of the footnote for a better reading experience. Showing the content of the footnote takes up more space and seeing which footnotes are identical references spotting `This is an example foot...` is harder versus seeing two `[1]` in the payload admin article.

tadjh: Just recieved confirmation from Jacob that he prefers numbered inlineBlocks over truncated notes. Pushing a fix on this PR

TedTimbrell:
Closing in favor of the other option

PR #616, opened 2026-04-27
Footnote Numerical Labels & Shortcut Feature

tadjh:
## Context

Jacob confirmed that numerical footnotes are preferred.

Branched off @TedTimbrell PR

- Restored numbered footnotes
- FootnoteLabels now update in realtime
- Optimized collectFootnotes (Only walks tree recursively a single time instead of twice)
- Removes lexical packages in favor of payload re-exports
- Optimize shortcut feature
- Fix shortcut options order

## Test Plan

- [ ] Solves stale `Footnote` label  (Footnotes now update on Autosave, instead of only on editor's save/publish)
- [ ] Insertion of copied footnotes is faster (shortcut feature for quick inserts)
- [ ] FootnotesList shows up in admin article view for writers while article status is still `draft` (Footnotes now recompute with each Autosave)
- [ ] FootnotesList shows up in Live Preview while article status is still in `draft` (Footnotes now recompute with each Autosave)

### DIFF
--- a/src/blocks/Footnote/FootnoteLabelClient.tsx
+++ b/src/blocks/Footnote/FootnoteLabelClient.tsx
@@ -17,7 +17,7 @@ export const FootnoteLabelClient: React.FC<FootnoteLabelClientProps> = ({ siblin
   const { data } = useDocumentInfo()
   const footnotes = (data?.footnotes as FootnotesField) ?? []
   const currentIndex =
-    index ?? (note ? (footnotes.find((f) => f.note === note)?.index ?? null) : null)
+    (note ? (footnotes.find((f) => f.note === note)?.index ?? null) : null) ?? index ?? null
 
   if (!note) return <span>Footnote</span>
   if (typeof currentIndex !== "number") return <span>{truncate(note, PREVIEW_LIMIT)}</span>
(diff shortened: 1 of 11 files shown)


src/blocks/Footnote/FootnoteLabelClient.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface FootnoteLabelClientProps {
+  siblingData?: Partial<FootnoteBlock> | null
+}
+
+export const FootnoteLabelClient: React.FC<FootnoteLabelClientProps> = ({ siblingData }) => {
+  const { note, sourceId, index } = siblingData ?? {}
+  const { data } = useDocumentInfo()
+  const footnotes = (data?.footnotes as FootnotesField) ?? []
+  const currentIndex =
+    (note ? (footnotes.find((f) => f.note === note)?.index ?? null) : null) ?? index ?? null
```
TedTimbrell, line 20: I do wonder if we're going to hit some issue at some where we have equivalent notes but separate attributions. Seems fine for this PR given we're more or less already doing this

TedTimbrell (approved):

src/blocks/Footnote/FootnoteLabelClient.tsx
tadjh: yeah I do worry about that as well, a possible edge case, but doubt we'll hit it within the same document, something to think about in future!

PR #620, opened 2026-04-29
Contributors Block

### LINKED ISSUE #608: Create About Page
- [ ] Create About Page
- [ ] List Chief Editors
- [ ] List Editors
- [ ] List Authors
- [ ] List Narrators (new user role)
- [ ] Add a new `checkbox` field to `Users` collection called `featured`. Featured will be used to surface users that we want to feature on the about page. `where : { featured: true }`.

We may need a block that can query collections with custom filters like users by role.

- [ ] Create Collection List Block

otherwise we'll have to hardcode the about page.

TedTimbrell:
## Context
Closes #608 

Screenshot 2026-04-29 at 7 32 36 PM

### DIFF
--- a/src/blocks/Content/Component.tsx
+++ b/src/blocks/Content/Component.tsx
@@ -17,7 +17,7 @@ export const ContentBlock: React.FC<ContentBlockProps> = (props) => {
   }
 
   return (
-    <div className="mx-auto my-4 max-w-3xl px-4">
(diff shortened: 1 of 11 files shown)


tadjh (approved):
I don't like that we changed all Content blocks width, but it is a good patch for now until we decide inner page layouts.

PR #621, opened 2026-04-29
Adjust Anchor Links for Sticky Header

TedTimbrell:
## Context

Clicking on a footnote link would smooth scroll you down but the footnote would be covered by the header. 

This adds an offset to clear the header.


tadjh (approved):
Good catch!

PR #622, opened 2026-04-29
Replace 'Articles in this volume' with volume title.

### LINKED ISSUE #481: Author Pages should show list of Volumes they've acted as Editor for
In author view such as `https://staging.pragmaticpapers.com/authors/teaganwordsmith`

Note: try to reuse the existing volumes query data that's already on each author slug page, and filter a copy of the volumes array for those with a matching author.

Reuse ArticleList by broading it's type to accept Articles and Volumes. Use a typegaurd (isArticle) if necessary.

### LINKED ISSUE #482: Volume single pages don't show Volume Titles
In this example view `https://staging.pragmaticpapers.com/volumes/3`
The volume title should be shown.

espdesign:
## Context
Closes https://github.com/digitalgroundgame/pragmatic-papers/issues/482
Summary
- Replaces the generic "Articles in this volume" heading with the actual volume title on volume pages falling back to original if missing volume title. (src/app/(frontend)/volumes/[slug]/page.tsx) 
- This improves UX by showing the specific volume name instead of a generic label

Changes
- src/app/(frontend)/volumes/[slug]/page.tsx — Updated to display volume title instead of static "Articles in this volume" text with fallback to original.

## Test Plan

1. Inspect src/app/(frontend)/volumes/[slug]/page.tsx: confirm static "Articles in this volume" is replaced with dynamic volume.title, and test fallback.
2. Test preview mode for volume page to confirm volume title renders correctly.


espdesign:
I made a design decision let me know what you think about it. @tadjh @TedTimbrell

tadjh (approved):
This makes sense to me and looks good. Only thing to check is how title field is used in production. I want them to use the titles that they use on social media for these volumes on the website. But right now I think it's usually all article titles or `Volume 3`. Which either will be less than ideal. But that's really debt we inherited from the old homepage design requirements.

espdesign:
> This makes sense to me and looks good. Only thing to check is how title field is used in production. I want them to use the titles that they use on social media for these volumes on the website. But right now I think it's usually all article titles or `Volume 3`. Which either will be less than ideal. But that's really debt we inherited from the old homepage design requirements.

I think this change will gently nudge them into the direction you would prefer.

tadjh:
Before we go thru this change I'd like to use the old Reddit titles to document accurate Volume Titles so that we can go thru the old Volumes and update their titles. I don't want to display the 7 title concat titles on the website.
 
 We can find the original volume titles on the reddit: https://www.reddit.com/r/Destiny/search/?q=Pragmatic+Papers

espdesign:
> Before we go thru this change I'd like to use the old Reddit titles to document accurate Volume Titles so that we can go thru the old Volumes and update their titles. I don't want to display the 7 title concat titles on the website.
> 
> We can find the original volume titles on the reddit: https://www.reddit.com/r/Destiny/search/?q=Pragmatic+Papers

I don't think the Reddit post title would be a good fit. This seems like something we should consult the pragmatic paper editors on. Perhaps even request they preform the task for us, then we can go in and change them.

tadjh:
> > Before we go thru this change I'd like to use the old Reddit titles to document accurate Volume Titles so that we can go thru the old Volumes and update their titles. I don't want to display the 7 title concat titles on the website.
> > 
> > We can find the original volume titles on the reddit: https://www.reddit.com/r/Destiny/search/?q=Pragmatic+Papers
> 
> I don't think the Reddit post title would be a good fit. This seems like something we should consult the pragmatic paper editors on. Perhaps even request they preform the task for us, then we can go in and change them. 

If you look at the images, the reddit posts have titles for each volume like "100 Days In". These are the true Volume titles. Ultimately it is up to the editors, but these titles were chosen by the previous editors and would like to retain their choices instead of making of names retroactively if we can avoid it.

tadjh:
Closing this until Editors are ready for this change. It's a good change but they aren't using this functionality yet.

PR #623, opened 2026-04-29
bugfix populateAuthors & add payload logging.

### LINKED ISSUE #574: Fix populatedAuthors hook: N+1 queries, silent error swallowing, missing virtual:true
## Problem

The `populateAuthors` afterRead hook in `src/collections/Articles/hooks/populateAuthors.ts` has several issues that should be addressed.

### 1. N+1 query pattern

The hook loops through each author and fires a separate `findByID` per author:

```ts
for (const author of doc.authors) {
  const authorDoc = await payload.findByID({ id: ..., collection: "users" })
}
```

This means an article with 3 authors costs 3 sequential DB round-trips on every read. It should use a single `payload.find` with `where: { id: { in: authorIds } }`.

### 2. Silent error swallowing

```ts
} catch {
  // swallow error
}
```

Failures are completely hidden. At minimum errors should be logged so problems are visible during development and in production observability tooling.

### 3. `populatedAuthors` field is missing `virtual: true`

The field is defined without `virtual: true`, which means Payload creates a database column for it even though the data is never persisted — it's always overwritten at read time by the hook.

### 4. `id` field typed as `text` with a cast

The `id` sub-field in the `populatedAuthors` array is `type: "text"`, but `User.id` is a `number`. This mismatch is papered over with `id: populatedAuthor.id as unknown as string`. The field type should match the actual ID type.

## Fix

- Replace the `for` loop with a single `payload.find({ collection: "users", where: { id: { in: authorIds } }, overrideAccess: true })`
- Add `virtual: true` to the `populatedAuthors` field definition
- Change the `id` sub-field to `type: "number"` and remove the cast
- Replace the bare `catch` with at minimum a `payload.logger.error(...)` call

espdesign:
## Context
Closes https://github.com/digitalgroundgame/pragmatic-papers/issues/574

Changes made:
1. Adding logging 
2. src/collections/Articles/hooks/populateAuthors.ts — Replaced the N+1 for loop with a single payload.find({ where: { id: { in: authorIds } } }), preserved author order with a Map lookup, added payload.logger.error() with authorIds context, removed the as unknown as string cast (now id: populatedAuthor.id directly)
3. src/collections/Articles/index.ts — Added virtual: true to the populatedAuthors field, changed the id sub-field from type: "text" to type: "number"
4. src/payload-types.ts — Regenerated via payload generate:types
5. src/app/(frontend)/authors/page.tsx — Removed String() cast on doc.id (now number)
6. src/app/(frontend)/volumes/[slug]/page.tsx — Changed Set<string> to Set<number> to match the updated id type

6. generated migration to virtualize the populatedAuthors field.
> #opencode
> The populatedAuthors field is always overwritten on every Article read by the populateAuthors afterRead hook, so persisting it to the database is redundant. Switching to virtual: true solves 3 problems:
> 1. Eliminates wasted DB storage: Payload stops creating/maintaining the articles_populated_authors and related tables (dropped in migration 20260429_190436.ts), since the data is never stored long-term.
> 2. Prevents stale data: If the field were persisted, saved copies could become outdated when underlying User data changes (e.g., a user updates their name or profile image). Virtual ensures you always get fresh data re-fetched by the hook on every read.
> 3. Simplifies the DB schema: Removes unnecessary tables, indexes, and foreign key constraints, reducing database overhead.

## Test Plan

Manual / API testing:

- [ ] 1. Single query behavior — Check server logs or DB query logs to confirm only 1 find call is made per article read (not N per author)
- [ ] 2. Author order preserved — Create an article with authors A, B, C, verify populatedAuthors comes back in that same order (not DB order)
- [ ] 3. Mixed author types — Test with doc.authors containing both raw number IDs and partially-populated User objects (simulates depth>0 upstream queries)
- [ ] 4. Error logging — Temporarily break the payload.find call, verify payload.logger.error fires with { err, authorIds } and the hook doesn't crash
- [ ] 5. depth: 1 — Verify profileImage and socials are populated in the response (not just IDs)
- [ ] 6. Virtual field — After applying migration 20260429_190436.ts, confirm articles_populated_authors tables are gone and the field still populates on read
- [ ] 7. Frontend rendering — Visit an article page, verify author bylines render correctly in ArticleHero + AuthorList
- [ ] 8. RSS feed + structured data — Check /feed.articles and view page source for correct author fields in JSON-LD


package.json
```diff
@@ -14,6 +14,7 @@
     "dev:db-down": "docker compose -p pragmatic-papers -f docker-compose.yml down",
     "dev:next": "cross-env NODE_OPTIONS=--no-deprecation next dev -p 8000",
     "dev:prod": "cross-env NODE_OPTIONS=--no-deprecation rm -rf .next && pnpm build && pnpm start",
+    "dev:pretty": "pnpm dev | pino-pretty",
```
TedTimbrell, line 17: 🙌

src/collections/Articles/hooks/populateAuthors.ts
```diff
@@ -12,35 +12,51 @@ export const populateAuthors: CollectionAfterReadHook<Article> = async ({
...
-      })
-
-      if (authorDoc) {
-        populatedAuthors.push(authorDoc)
-      }
-    } catch {
-      // swallow error
+  const authorIds = doc.authors.map((author) =>
+    typeof author === "number" ? author : author?.id,
+  ) as number[]
```
TedTimbrell:
```suggestion
  )
```

TedTimbrell: I think the cast is unnecessary

TedTimbrell (dismissed):
LGTM

tadjh (commented):
Looks good code wise! Idk what `pino-pretty` is or why we need these logs. But I saw @TedTimbrell reacted with 🙌 so I assume this is wanted? But otherwise good stuff!

espdesign:
>  🙌

Pino-pretty turns the json log (payload uses pino)to pretty log 😍. Logging will be useful in the long run imo.

TedTimbrell (approved):

tadjh (approved):

PR #629, opened 2026-05-02
General Refactoring and Deduplication

yunidev-uk:
I'd noticed there's a decent bit of code, particularly in helpers, that seems to be either duplicated or overly verbose. This PR refactors a bunch of stuff to try and simplify and clean up some of this.

- `access` is now just a single 54 line `index.ts` and is more organised, instead of a bunch of file with 5-10 lines each
- Querying of articles, volumes, users and such has been moved into helpers to allow reuse, deduplicating some code
- Removal of dead code
- Creating helpers for functionality that was widely duplicated
  - Page number parsing
  - Reused configs
  - Hooks for change/create events on various collections
- Some other minor refactoring, e.g. `toRoman` function, cleaning up some incorrect naming, helpers for defining image sizes, and more

I've not had a particularly deep look at `components` - from a glance it seems there are a fair few unused or near-duplicated components but it's possible there are plans for them, so I've opted to not clean them up.

PR #630, opened 2026-05-03
Refactor access helpers

yunidev-uk:
This was originally included as part of #629, but was asked to be extracted out into a separate PR to allow testing for regressions to be done more easily.

In terms of changes, this PR merges all access management helpers into a single organised file and reduces duplication.


tadjh:
Look good at first glance. Will try this PR out over next few days to see if there's any edge cases that we should handle.

espdesign (approved):
LGTM

PR #639, opened 2026-05-05
(test)Fix: Forced Image Downloads via Domain Isolation

espdesign:
## Context
I don't enjoy the forced attachment downloads on production. However they are used by default for security purposes. 

## Test Plan
How to Verify
1.  Inspect an Image: Open the inspector and verify the  tag src points directly to Supabase/S3 (or /media/ in local dev) rather than /_next/image.
2.  Check Headers: In the Network tab, verify the image request no longer contains the X-Nextjs-Cache header, confirming it is bypassing the proxy.
3.  View Inline: Right-click a WebP image and "Open in new tab." It should display in the browser.

## Question:
_Jarvis_ is telling me that its currently using a redundant cache, where /_next/image cache stored re-optimized versions of images, which may be unnecessary because payload already pre-processing images into sized variants and webp format. and a possible  deduced Server Load. Because our Next.js server no longer needs to spend CPU cycles re-processing images that have already been optimized by the CMS. 

Any ideas of the validity of these statements?

### DIFF
--- a/src/components/Media/ImageMedia/index.tsx
+++ b/src/components/Media/ImageMedia/index.tsx
@@ -62,6 +62,7 @@ export const ImageMedia: React.FC<ImageMediaProps> = ({
   loading,
   fill,
   style,
+  unoptimized = true,
   ...props
 }) => {
   if (!media.url) return null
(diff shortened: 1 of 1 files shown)


tadjh:
I think Jarvis is tripping. I think we still want to optimize our image delivery further not less. But could be overkill. Need to research more. If we do this I expect our lighthouse score to drop as a result.

PR #640, opened 2026-05-07
Client Side Recommendation

TedTimbrell:
## Context

The server side rendered recommendations limit us to a single result set that's cached for everyone (or every page).

This PR
* Moves it to a client component
* Sources the top rec list from the server
    * Server response is cached in CF returns all top 20 articles and updates once every ten minutes
* Uses weighted sampling to try to distribute views across articles. Handled on the client side 
* Adds google analytics events so we can track CTR


tadjh (approved):
LGTM

PR #641, opened 2026-05-07
try dynamic imports for rarely used components to decrease average bu…

TedTimbrell:
…ndle size


tadjh:
We've made major refactors since this PR to address some of these bundle issues. Need to rerun the bundle analyzer after merging the conflicts.

PR #653, opened 2026-05-13
Add env-aware install wrapper

evan6seven:
## Summary

- Adds `pnpm install:env` for local installs that need values from `.env`.
- Moves `.env` creation out of the install lifecycle so production and CI installs do not copy local env files.
- Updates onboarding docs to keep the private font token in the project `.env` instead of a global machine environment variable.


evan6seven:
Will address with #683

PR #655, opened 2026-05-14
Let users specify their own envrc

TedTimbrell:


tadjh:
@TedTimbrell Does this has any impact on coolify, or just local?

tadjh (approved):
LGTM

TedTimbrell:
> @TedTimbrell Does this has any impact on coolify, or just local?

just local

PR #667, opened 2026-05-15
Stop database container when exiting dev

evan6seven:
## Before
Postgres database stays running on your system all the time

## After
Press Ctrl C to exit the dev server and stop the database


espdesign:
Yeah we had this but I think issues with windows running both. @tadjh can you confirm on your end again?

tadjh:
> Yeah we had this but I think issues with windows running both. @tadjh can you confirm on your end again?

I'll try it again

evan6seven:
Will address in another PR

PR #680, opened 2026-05-15
[Testing] Extend Search with tsvector

TedTimbrell:
Replace LIKE-on-title with websearch_to_tsquery + ts_rank_cd against a generated tsvector column that weights title > authors > excerpt > body. beforeSync now flattens Article Lexical content into a hidden `body` field so full article text actually makes it into the search index.


tadjh:
This is working! @espdesign thoughts?

tadjh (approved):

tadjh (approved):

PR #682, opened 2026-05-15
chore(docker): improve build caching

espdesign:
#  Dockerfile Dependency Caching Optimization

## 📈 Performance Results

Based on definitive testing in Coolify for a simple source code change (e.g., modifying a React component):

- **Old Dockerfile (Baseline)**: 3m 38s
- **New Optimized Dockerfile**: **2m 07s**
- **Time Saved**: **1m 31s (~42% faster on every code change)**

## 🚀 Overview

This PR optimizes the Docker build process by implementing a multi-stage `pnpm` caching strategy that intelligently handles different types of codebase changes.

## 🛠 Key Improvements
1.  **Global Store Caching**: Both `pnpm fetch` and `pnpm install` now utilize a shared Docker BuildKit cache mount for the global store (`/pnpm/store`). This ensures that dependencies are preserved across builds even when the project layers are invalidated.
2.  **Lockfile-First Fetching**: The `pnpm fetch` step now depends strictly on `pnpm-lock.yaml` and `.npmrc`. This separates the expensive "download" phase from `package.json` changes.
3.  **Surgical Script Injection**: Only the scripts required for `postinstall` (e.g., font installation) are copied before the dependency installation. This prevents changes to unrelated scripts (like test runners or migration helpers) from busting the dependency cache.
4.  **Offline Installation**: The final `pnpm install` runs in `--offline` mode. It links dependencies directly from the cached global store, making the installation phase extremely fast.
5.  **Version Synchronization**: A bind mount is used during the base stage to read `package.json` and ensure the `pnpm` version used during fetching perfectly matches your local pinned version, preventing cache misses due to version drift.
6.  **Next.js Build Caching**: Added a dedicated BuildKit cache mount for `.next/cache`. This allows Next.js to reuse previously compiled chunks and telemetry data, significantly accelerating the compilation phase.
7.  **Build Phase Markers**: Added explicit `--- PHASE: ... ---` log markers in the Dockerfile to provide high-visibility feedback in Coolify logs.

## 📝 Verification

- Verified that `pnpm fetch` correctly utilizes the cache mount.
- Validated that `HUSKY=0` is set to skip unnecessary git hooks during build.
- Confirmed that source-code only changes successfully skip the dependency installation phase entirely.


tadjh (dismissed):
LGTM! Only thing is make sure `/app/` isn't referenced in production anywhere since it seems we are now directly on the root instead of a nested folder, which is fine, just making sure.

dockerfiles/PragmaticPapers.Dockerfile
```diff
@@ -91,101 +97,60 @@ ENV NEXT_PUBLIC_GOOGLE_ANALYTICS_ID=${NEXT_PUBLIC_GOOGLE_ANALYTICS_ID}
...
-# Build application with migrations
-# Runs migrations and then builds
-# Source the potentially modified DATABASE_URI before building
-RUN . /tmp/build.env && \
-    pnpm install --frozen-lockfile && \
+    if [ -f /tmp/database_uri.env ]; then . /tmp/database_uri.env; fi && \
+    /usr/local/bin/copy-database.sh && \
+    echo "--- PHASE: DATABASE MIGRATIONS ---" && \
+    export DATABASE_URI=$DATABASE_URI && \
     pnpm payload migrate && \
```
TedTimbrell, line 115: tangent: we really need to move migrations out of the docker file

dockerfiles/PragmaticPapers.Dockerfile
```diff
@@ -91,101 +97,60 @@ ENV NEXT_PUBLIC_GOOGLE_ANALYTICS_ID=${NEXT_PUBLIC_GOOGLE_ANALYTICS_ID}
...
-
-# Build application with migrations
-# Runs migrations and then builds
-# Source the potentially modified DATABASE_URI before building
-RUN . /tmp/build.env && \
-    pnpm install --frozen-lockfile && \
+    if [ -f /tmp/database_uri.env ]; then . /tmp/database_uri.env; fi && \
+    /usr/local/bin/copy-database.sh && \
+    echo "--- PHASE: DATABASE MIGRATIONS ---" && \
+    export DATABASE_URI=$DATABASE_URI && \
```
TedTimbrell: Is this export still needed?

dockerfiles/scripts/start.sh
```diff
@@ -0,0 +1,38 @@
...
+echo "Environment: $NODE_ENV"
+echo "Database: PostgreSQL"
+echo "Port: $PORT"
+echo "Hostname: $HOSTNAME"
+echo "Storage: $([ "$USE_LOCAL_STORAGE" = "true" ] && echo "Local" || echo "S3")"
+echo "========================================="
+
+# 1. Basic Database Connectivity Check
+# This ensures the DB is reachable before Node starts, but we don't run migrations here.
+# Migrations are now handled in the build stage.
```
TedTimbrell: rm, this comment is accurate to earlier in this branch's commit history but not dev/main (afaict).

dockerfiles/scripts/start.sh
```diff
@@ -0,0 +1,38 @@
...
+echo "Verifying database reachability..."
+# Extract host and port from DATABASE_URI (primitive sh parser)
+DB_HOST=$(echo $DATABASE_URI | sed -e 's|.*@||' -e 's|/.*||' -e 's|:.*||')
+DB_PORT=$(echo $DATABASE_URI | sed -e 's|.*:||' -e 's|/.*||')
```
TedTimbrell: nit: if someone forgets to put in a port this ends up logging the password.

dockerfiles/scripts/start.sh
```diff
@@ -0,0 +1,38 @@
...
+
+MAX_RETRIES=10
+RETRY_COUNT=0
+
+until nc -z "$DB_HOST" "$DB_PORT" || [ $RETRY_COUNT -eq $MAX_RETRIES ]; do
+  RETRY_COUNT=$((RETRY_COUNT+1))
+  echo "Waiting for database at $DB_HOST:$DB_PORT... ($RETRY_COUNT/$MAX_RETRIES)"
+  sleep 3
+done
+
```
TedTimbrell: What's the purpose of this waiting on the database in this way? I'm not against it necessarily

TedTimbrell (commented):
lgtm

dockerfiles/PragmaticPapers.Dockerfile
```diff
@@ -3,189 +3,159 @@
...
-    chown nextjs:nodejs /app/start.sh
-
-# Switch to non-root user
+# Prepare media directory and set permissions
+RUN mkdir -p public/media \
+    && chown -R nextjs:nodejs . \
+    && chmod -R 755 public/media
+
+# Startup script configuration
+COPY --from=builder --chown=nextjs:nodejs /app/dockerfiles/scripts/start.sh ./start.sh
```
TedTimbrell, line 153: 🙌

dockerfiles/PragmaticPapers.Dockerfile
espdesign: Wouldn't that require payload and deps in production?

dockerfiles/PragmaticPapers.Dockerfile
espdesign: I think its needed for the copy-database to handle the correct url.

dockerfiles/scripts/start.sh
espdesign: True will fix.

dockerfiles/scripts/start.sh
espdesign: Should we just remove the echo for simplicity?

dockerfiles/scripts/start.sh
espdesign: It should not be needed if the database is live always.

dockerfiles/PragmaticPapers.Dockerfile
TedTimbrell: I don't entirely follow what added dependencies we'd need but yeah I think the final output image should be able to manage migrations running and is worth whatever increase in image size it might cause. 

Having it during build works but feels like a potentially nasty footgun.

dockerfiles/scripts/start.sh
TedTimbrell: could toss it to node and use a URL parse? That or remove it I think?

dockerfiles/PragmaticPapers.Dockerfile
TedTimbrell: Sorry, just to clarify, we'd add the command here (versus having it run in each instance's startup):

Screenshot 2026-05-17 at 1 41 16 PM

dockerfiles/PragmaticPapers.Dockerfile
TedTimbrell: `modify-database-uri.sh` I think is doing it?

dockerfiles/PragmaticPapers.Dockerfile
espdesign: I don't feel like chasing down that rabbit-hole again tbh. I do agree tho.

TedTimbrell: 😆 not here in this PR ofc.

TedTimbrell (approved):

PR #683, opened 2026-05-15
Run dev server in Docker Compose

evan6seven:
## Summary

- Run the Next.js dev server inside Docker Compose alongside Postgres.
- Reuse the production Dockerfile base stage for the dev container Node image.
- Keep Postgres private on an internal Docker network while exposing only the app on port 8000.
- Update onboarding docs for the Docker-based font install flow.
- Github CI has its own env defaults


evan6seven:
Updated so that pnpm isn't downloaded every time you start the container.

espdesign (dismissed):
LGTM

espdesign (approved):
LGTM

PR #686, opened 2026-05-16
Restore local development and pre-push configurations

espdesign:
## Context

Removes the Next.js app from Docker Compose so developers run it directly on the host (faster iteration, simpler debugging, better hot reload). PostgreSQL stays in Docker.

**Docker simplification:**
- Removed `payload-pragmatic` service (Next.js container) from `docker-compose.yml` — it now only manages PostgreSQL
- Exposed Postgres on port 9000 (host) instead of an internal Docker network
- Removed `docker-husky.sh` — Husky hooks now run directly on the host
- Removed unused Docker volumes (node_modules, pnpm store) and networks

**Scripts & tooling:**
- Split `pnpm dev` into `dev:db` (Docker Postgres) + `dev:next` (Next.js on host with Turbopack)
- `dev:db-nuke` now uses `payload migrate:fresh` instead of Docker volume removal
- `install-fonts.mjs` auto-creates `.env` from `.env.example` on first install
- Updated `devenv.nix` to enable dotenv and pin Node.js 24
- Updated `.gitignore` for devenv/direnv artifacts

**CI:**
- Added Playwright browser cache to speed up E2E workflow
- Playwright `webServer.command` now points to `dev:next` (avoids Docker in CI)
- Removed unused `restore-nextjs-cache` input from `setup-project` action

**Docs & config:**
- Updated `README.md` dev instructions, font token setup guidance, and script descriptions
- Fixed `DATABASE_URI` in `.env.example` to use `localhost:9000`
- Restored `.npmrc` `GH_FONT_READ` variable syntax

## Test Plan

1. **Clean start:** Run `pnpm dev` — verify PostgreSQL starts via Docker and Next.js dev server starts on port 8000
2. **Dev loop:** Edit a file and verify hot reload works via Turbopack (sub-second refresh)
3. **Husky hooks:** Stage a file and commit — verify `lint-staged` runs directly on host (not in Docker)
4. **Push:** Run `git push` — verify `pnpm check-types` runs directly on host
5. **Database reset:** Run `pnpm dev:db-nuke` — verify it re-runs migrations instead of requiring Docker volume removal
6. **Font install:** Delete `public/fonts/` and `node_modules/`, run `pnpm install` — verify fonts are copied and `.env` is auto-created
7. **E2E tests:** Run `pnpm test:e2e` — verify Playwright starts Next.js directly (not via Docker)
8. **CI:** Verify Playwright workflow passes with browser caching enabled


tadjh:
I'm experiencing longer wait times in development due to the webserver being ran in docker. Tests slower as well. Alot of waiting around for docker to start and stop. Also when the next webserver crashes I have to restart both database and webserver, when only webserver needs restarted.

I'm team revert the change to have everything in docker. We should save that for production builds only. Thinking int/e2e should use testcontainers.

package.json
tadjh, line 1: we need node 24 going forward

espdesign: strictly? it has a WARNING when its set high on my box.

.npmrc
tadjh, line 1: Revert this. This surpresses GH_FONT_READ warnings

scripts/install-fonts.mjs
tadjh, line 1: We removed copy-env from `preinstall` so everyone is now in charge of creating their own .env file. Do we need this to make install-fonts work now?

tadjh (dismissed):

package.json
```diff
@@ -8,10 +8,12 @@
     "postinstall": "node scripts/install-fonts.mjs",
     "build": "cross-env NODE_OPTIONS=--no-deprecation next build",
     "postbuild": "next-sitemap --config next-sitemap.config.cjs",
-    "dev:db-nuke": "docker compose -p pragmatic-papers -f docker-compose.yml down && docker volume rm pragmatic_papers_postgres_data",
+    "dev": "pnpm dev:db && pnpm dev:next",
```
evan6seven: single change to stop the db when we are done:

`pnpm dev:db && pnpm dev:next && pnpm dev:db-down`

tadjh: this syntax doesnt work. && will run all three, oh you're still trying to close db after. I'll have to switch commands, but lemme try this and see how it feels.

espdesign:
do me a favor and run the e2e locally, see if thats working.

package.json
tadjh: after testing it feels good

tadjh (approved):

PR #694, opened 2026-05-18
Fix search body overflow for long articles

### LINKED ISSUE #695: Search sync fails for long articles: body field exceeds 40000 char limit
## Description

Saving or publishing a long article causes the search sync to fail with a 400 error. The `body` field on the `search` collection is a `textarea`, which Payload enforces a 40000-character max on at the application level. Long articles exceed this when their Lexical content is converted to plain text.

## Error

```
{"level":50,...,"err":{"data":{"collection":"search","errors":[{"label":"Body","message":"This value must be shorter than the max length of 40000 characters.","path":"body"}]},"status":400},"msg":"Error syncing search document related to articles with id: '241'."}
```

## Cause

`beforeSync` in `src/plugins/index.ts` extracts the full plain-text body from Lexical content and passes it to the search collection without length bounds. The `textarea` field type has a default `maxLength` of 40000 enforced by Payload validation.

## Affected file

`src/plugins/index.ts` — `beforeSync` hook

tadjh:
Closes #695

## Summary

- The `body` field on the `search` collection is a `textarea` with a 40000-character Payload validation limit. Long articles (e.g. id 241) exceeded this on every save, logging a 400 error and failing to sync to the search index.
- Fixed by truncating the plain-text body to 39000 chars in the `beforeSync` hook (`src/plugins/index.ts`).

## Test plan

- [ ] Publish or re-save a long article and confirm no search sync error in logs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,9 @@
 name: CI
 
 on:
+  push:
+    branches: [main]
   pull_request:
-    branches: [main, dev]
 
 jobs:
   static:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, dev]
 
 jobs:
   test:
--- a/tests/integration/search-fts.test.ts
+++ b/tests/integration/search-fts.test.ts
@@ -98,7 +98,6 @@ describe("Full-Text Search Integration", () => {
       AND id = ${searchDoc.id}
     `)
 
-    // If the row is found, the FTS index correctly contains the keyword
     expect(ftsResult.rows.length).toBe(1)
   })
 })
--- a/tests/integration/search-fts.test.ts
+++ b/tests/integration/search-fts.test.ts
@@ -1,0 +1,104 @@
+import { getPayloadConfig } from "@/utilities/getPayloadConfig"
+import type { Payload } from "payload"
+import { afterAll, beforeAll, describe, expect, it } from "vitest"
+import { sql } from "@payloadcms/db-postgres/drizzle"
+
+// Shape of a row returned by the raw FTS query below
+interface SearchResultRow {
+  id: string | number
+  body?: string
+}
+
+// Minimal interface for the Drizzle query executor exposed by Payload's DB adapter
+interface DrizzleExecutor {
+  execute: (query: ReturnType<typeof sql>) => Promise<{ rows: SearchResultRow[] }>
+}
+
+// The Payload DB adapter carries a `drizzle` property we need to run raw SQL.
+// We cast to this interface because the public Payload type doesn't expose it.
+interface DatabaseAdapterWithDrizzle {
+  drizzle: DrizzleExecutor
+}
+
+describe("Full-Text Search Integration", () => {
+  let payload: Payload
+
+  beforeAll(async () => {
+    payload = await getPayloadConfig()
+  })
+
+  afterAll(async () => {
+    // Clean up the database connection so Testcontainers can shut down cleanly
+    await payload.db.destroy?.()
+  })
+
+  it("truncates and indexes long body text for FTS searchability", async () => {
+    // Unique keyword we will search for later — must not appear in other test data
+    const keyword = "polyglotism"
+    const articleTitle = "Extremely Long Article Test"
+
+    // Create a published article whose Lexical content is one paragraph containing
+    // the keyword repeated 10 000 times. This produces a body far larger than the
+    // 39 000-character truncation limit, proving the hook truncates before indexing.
+    const _article = await payload.create({
+      collection: "articles",
+      context: { disableRevalidate: true }, // skip ISR revalidation during tests
+      data: {
+        title: articleTitle,
+        _status: "published",
+        slug: "long-article-test",
+        content: {
+          root: {
+            type: "root",
+            format: "",
+            indent: 0,
+            version: 1,
+            direction: "ltr",
+            children: [
+              {
+                type: "paragraph",
+                format: "",
+                indent: 0,
+                version: 1,
+                direction: "ltr",
+                children: [{ text: `${keyword} `.repeat(10000), type: "text", version: 1 }],
+              },
+            ],
+          },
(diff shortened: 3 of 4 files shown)


tadjh:
This truncation is one potential fix which has the consequence of biasing search results to the first 40k characters. Seems like a fine compromise.

The another fix is to see what our largest articles max lengths are currently, when ran thru the toPlaintext function that serch uses and set `textarea` `maxLength` out of range of this error.

Fee free to take over this PR.

src/jobs/updateRecommendations/logic.ts
```diff
@@ -61,7 +61,11 @@ export function readGA4Env(): {
 } {
   const propertyId = process.env.GA4_PROPERTY_ID
   const clientEmail = process.env.GOOGLE_SERVICE_ACCOUNT_EMAIL
+  // Some deployment platforms double-escape backslashes in env vars, producing
+  // \\n (three chars) instead of \n (two chars). Replace both levels.
   const privateKey = process.env.GOOGLE_SERVICE_ACCOUNT_PRIVATE_KEY?.replace(/\\n/g, "\n")
+    ?.replace(/\\n/g, "\n")
+    ?.trim()
```
TedTimbrell: Can we split this out into its own PR. I know that staging and dev deployments are working.

tadjh: Sure

espdesign:
> This truncation is one potential fix which has the consequence of biasing search results to the first 40k characters. Seems like a fine compromise.
> 
> The another fix is to see what our largest articles max lengths are currently, when ran thru the toPlaintext function that serch uses and set `textarea` `maxLength` out of range of this error.
> 
> Fee free to take over this PR.

truncation is perfect, we may want to truncate even more.

src/jobs/updateRecommendations/logic.ts
tadjh: #697

TedTimbrell (dismissed):

espdesign:
Possible test addition: https://github.com/digitalgroundgame/pragmatic-papers/pull/698

tadjh:
Added the tests in to this PR, need review.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,9 @@
 name: CI
 
 on:
+  push:
+    branches: [main]
   pull_request:
-    branches: [main, dev]
```
TedTimbrell, line 5: RM?

.github/workflows/playwright.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, dev]
```
TedTimbrell, line 7: same here?

tests/integration/search-fts.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    // `search_vector` tsvector column was populated and can match our keyword.
+    const db = (payload.db as unknown as DatabaseAdapterWithDrizzle).drizzle
+    const ftsResult = await db.execute(sql`
+      SELECT id 
+      FROM search 
+      WHERE search_vector @@ to_tsquery('english', ${keyword})
+      AND id = ${searchDoc.id}
+    `)
+
+    // If the row is found, the FTS index correctly contains the keyword
```
TedTimbrell: rm, explaining what rather than why

tests/integration/search-fts.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // Verify exactly one search document was created for our article
+    expect(searchDocs.length).toBe(1)
+    const searchDoc = searchDocs[0]!
+
+    // The body field must have been truncated to ≤ 39 000 characters so the
+    // Postgres `tsvector` column stays within the indexable size limit.
+    expect(searchDoc.body!.length).toBeLessThanOrEqual(39000)
+
+    // Bypass Payload's query layer and hit Postgres directly to confirm that the
```
TedTimbrell: This is to match what we're doing in production right? It would be good to mention that as it's somewhat ambiguous if this is a test thing

tests/integration/search-fts.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    // After creating the article, a hook should have synced it into `search`.
+    const { docs: searchDocs } = await payload.find({
+      collection: "search",
+      where: {
+        slug: { equals: "long-article-test" },
+      },
+      overrideAccess: true,
+    })
+
+    // Verify exactly one search document was created for our article
```
TedTimbrell: rm

tests/integration/search-fts.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    expect(searchDocs.length).toBe(1)
+    const searchDoc = searchDocs[0]!
+
+    // The body field must have been truncated to ≤ 39 000 characters so the
+    // Postgres `tsvector` column stays within the indexable size limit.
```
TedTimbrell: rm

TedTimbrell (commented):
props for the tests

.github/workflows/ci.yml
tadjh: This was preventing tests from running on PRs that are pointed towards another PR.

For example forks off PRs that still point to their fork, don't run these actions.  

If we only want tests on PRs pointed at dev and main then we can restore.

.github/workflows/playwright.yml
tadjh: This was preventing tests from running on PRs that are pointed towards another PR.

For example forks off PRs that still point to their fork, don't run these actions.  

If we only want tests on PRs pointed at dev and main then we can restore.

TedTimbrell: Oh my bad, I see. So excluding this param has it run on everything.

tadjh: yep that's what Evan said when he removed these, that they still run tests on PRs, just not limited to dev/main

tests/integration/search-fts.test.ts
tadjh: @espdesign

tadjh:
@espdesign Ted had a few notes left as comments on the new tests

PR #705, opened 2026-05-19
Unit/Integration Tests for Access Control Functions & Refactor of ACF

### LINKED ISSUE #661: Unit tests for access control functions
## Goal
Add unit tests for all access control functions in `src/access/`. These are pure functions that accept `{ req: { user } }` objects, making them ideal for isolated unit testing with mock user fixtures.

## Scope
- `anyone.ts` — always returns true
- `authenticated.ts` — returns Boolean(user)
- `authenticatedOrPublished.ts` — authenticated users OR published status
- `admins.ts` — delegates to isAdmin()
- `checkRole.ts` — isAdmin() helper
- `editor.ts` — role === 'editor'
- `editorOrSelf.ts` — editor OR own document
- `staff.ts` — role === 'staff'
- `writer.ts` — role === 'writer'
- `adminOrSelf.ts` — admin OR own document

## Test patterns
Create a shared mock user factory for admin, editor, writer, staff, and unauthenticated (null) users. Test each function with each user role permutation.

## Acceptance Criteria
- `src/access/__tests__/` contains test files for all 10 functions
- Each access function tested with all relevant user roles + unauthenticated case
- `pnpm test:unit` passes

espdesign:
### Summary

Migrates the user roles schema from a single-select `role` field to a multi-select `roles` field (`hasMany: true` in Payload) and refactors access control policies to support flat overlapping roles using a standardized naming convention: `is[Condition]Or[Role]`.

---

#### 1. Multiple User Roles
- **Schema Update**: Renamed `role` (select) to `roles` (select with `hasMany: true`) in `src/collections/Users/index.ts`. Default value set to `["member"]`.
- **Flat Checking Helpers**:
  - Implemented `hasRole(user, role | role[])` to perform pure role membership checks.
  - Implemented `hasRoleOrAdmin(user, role | role[])` to perform permission checks allowing admins/chief-editors to bypass.
  - Implemented `isAdmin(user)` and `isStaff(user)` (using `hasRoleOrAdmin` internally).
- **Explicit Access Checks**: Configured all access callbacks (`admin`, `editor`, `writer`, `staff`) to utilize these helpers flatly without hierarchical lists.

#### 2. Access Control Policies Standardized to `is[Condition]Or[Role]`
- Renamed and exported custom policies consistently across:
  - `isSelfOrAdmin` (allows user matching own record, admins exempt).
  - `isCreatedByOrEditor` (allows user who created the doc, editors/admins exempt).
  - `isDraftOrEditor` (restricts writers to draft documents they created, editors/admins exempt).
  - `isPublishedOrStaff` (allows staff to view drafts, others restricted to published).
- Updated JSDoc documentation at the top of `src/access/policies.ts`.
- Renamed integration test files and test suites under `tests/integration/access/` to align with the `Or` naming pattern (e.g. `isSelfOrAdmin.test.ts`).
- Renamed references in `AGENTS.md`.

#### 3. Test Coverage & Seeding
- Created a dedicated unit test suite for all access control policies in `src/access/__tests__/policies.test.ts`.
- Updated `roles.test.ts` to test pure role checks, explicit admin escalation, and overlapping roles.
- Updated `createUser` test helper and integration test assertions.
- Updated user seeding script (`src/endpoints/seed/users.ts`) to assign arrays of roles.

#### 4. Database Migration
- Created SQL migration script `src/migrations/20260523_010628_migrate_role_to_roles.ts` to safely copy existing users' single role values over to the new `users_roles` join table during migration, and map them back if rolled back.


tests/integration/access/narrator.test.ts
```diff
@@ -0,0 +1,235 @@
...
+      payload.update({
+        collection: "media",
+        id: media.id,
+        overrideAccess: false,
+        user: narrator,
+        context: { disableRevalidate: true },
+        data: { alt: "Should Not Update" },
+      }),
+    ).rejects.toThrow()
+  })
```
tadjh: Is a narrator allowed to edit (update) their own files?

src/payload.config.ts
```diff
@@ -87,6 +87,9 @@ export default buildConfig({
     pool: {
       connectionString: process.env.DATABASE_URI,
     },
+    // Disable auto-push in tests — migrations are run explicitly.
+    // Defaults to on in dev for convenience (schema syncs without manual migration).
+    push: process.env.PAYLOAD_PUSH_SCHEMA !== "false",
```
tadjh: Should this env var be aadded to .env.example or ci only?

src/access/__tests__/roles.test.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
TedTimbrell: rm, `any` is the devil. Could this instead use something to the effect of:
```
import type { User } from "@/payload-types"

const makeUser = (role: User["role"]): User => ({ role }) as User
const makeArgs = (user: User | null) => ({ req: { user } }) as Parameters<typeof staff>[0]
```

src/access/roles.ts
```diff
@@ -0,0 +1,62 @@
+import type { Access, AccessArgs, FieldAccess } from "payload"
+import type { User } from "@/payload-types"
+
+export type Role = NonNullable<User["role"]>
+
+export const ROLE_HIERARCHY: Record<Role, number> = {
```
TedTimbrell: This works for now but I think we'll want exclusive checks for each role at some point and a hierarchy pattern might get a bit confusing. 

e.g. the article-recommendations collection is bloat that developers might want to see but in an ideal world we'd want to disable for the chief-editor or editors.

src/access/__tests__/roles.test.ts
tadjh: Good catch

tadjh:
Do we take on #705 or #710, or both?

espdesign:
> Do we take on #705 or #710, or both?

I'm pretty happy with 710 especially policies. But I'm biased.

espdesign:
WIP: Flattened roles, the goal is to allow users to have multiple roles. Still need to work through the fine details.

src/plugins/index.ts
tadjh, line 1: This is what saves us from overflowing the textarea

tadjh (changes requested):

src/app/(frontend)/authors/[slug]/page.tsx
Comment (tadjh) [inline src/app/(frontend)/authors/[slug]/page.tsx:1]: Do we want admins to have generated path slugs  here?

src/app/(frontend)/authors/page.tsx
tadjh, line 1: Do we want to show admins on the frontend?

src/collections/Articles/index.ts
```diff
@@ -230,8 +228,8 @@ export const Articles: CollectionConfig = {
       hasMany: true,
       relationTo: "users",
       filterOptions: {
-        role: {
-          in: ["writer", "editor", "chief-editor"],
+        roles: {
+          in: ["writer", "editor", "chief-editor", "admin"],
```
tadjh: This will make it so admins are selectable as an article author?

src/collections/Media/index.ts
```diff
@@ -80,8 +79,8 @@ export const Media: CollectionConfig = {
       type: "relationship",
       relationTo: "users",
       filterOptions: {
-        role: {
-          equals: "narrator",
+        roles: {
+          in: ["narrator", "admin", "chief-editor"],
```
tadjh: Role change not needed, we'd simply add narrator role to a user, this way admins are showing up here by default

tadjh (changes requested):

dockerfiles/scripts/start.sh
tadjh, line 1: Does this belong on this PR?

tadjh (dismissed):

tadjh:
This is PR is in a good state but in no rush to run that database migration given the roles reassignment. Looking for a feature that depends on non-hierarchical user permissions before merging this in.

tadjh:
Filed #883 to track Cara's request that a single user be able to hold both the **Writer** and **Narrator** roles at once — directly enabled by the multi-role model introduced in this PR.

tadjh (approved):

PR #706, opened 2026-05-19
Media references with deletion control

espdesign:
Stacked PR targeting #705 — adds media references tracking and deletion control for published media.


tadjh (commented):
Ah is this so they don't accidently delete media that's already in use?

tadjh:
Not sure what this PR is targeting.

espdesign:
> Not sure what this PR is targeting.

Preventing media in use from being deleted while also clarifying to user why it cannot be deleted.

PR #708, opened 2026-05-20
Bugfix RSS Feed Links

TedTimbrell:
## Context

We're using the unencoded urls for the image links which for RSS readers simply fails. Thanks to Ky for flagging it.

### DIFF
--- a/src/utilities/generateRssFeed.ts
+++ b/src/utilities/generateRssFeed.ts
@@ -20,10 +20,13 @@ import { getServerSideURL } from "./getURL"
 const SITE_URL = getServerSideURL()
 
 const getMediaUrl = (url: string) => {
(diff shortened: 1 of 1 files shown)


tadjh (commented):
Seems good, what's a link I can visit to see the difference?

TedTimbrell:
> Seems good, what's a link I can visit to see the difference?

https://pr-708.pragmaticpapers.com/feed.volumes

Screenshot 2026-05-19 at 10 59 55 PM

You can see here how the image url is encoded. 

You can see [here](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fpragmaticpapers.com%2Ffeed.volumes) that the example of spaces in a media file name breaks the parsing

tadjh (approved):

PR #710, opened 2026-05-21
feat: support multiple user roles and standardize flat access policies

espdesign:
### Summary

Migrates the user roles schema from a single-select `role` field to a multi-select `roles` field (`hasMany: true` in Payload) and refactors access control policies to support flat overlapping roles using a standardized naming convention: `is[Condition]Or[Role]`.

---

#### 1. Multiple User Roles
- **Schema Update**: Renamed `role` (select) to `roles` (select with `hasMany: true`) in `src/collections/Users/index.ts`. Default value set to `["member"]`.
- **Flat Checking Helpers**:
  - Implemented `hasRole(user, role | role[])` to perform pure role membership checks.
  - Implemented `hasRoleOrAdmin(user, role | role[])` to perform permission checks allowing admins/chief-editors to bypass.
  - Implemented `isAdmin(user)` and `isStaff(user)` (using `hasRoleOrAdmin` internally).
- **Explicit Access Checks**: Configured all access callbacks (`admin`, `editor`, `writer`, `staff`) to utilize these helpers flatly without hierarchical lists.

#### 2. Access Control Policies Standardized to `is[Condition]Or[Role]`
- Renamed and exported custom policies consistently across:
  - `isSelfOrAdmin` (allows user matching own record, admins exempt).
  - `isCreatedByOrEditor` (allows user who created the doc, editors/admins exempt).
  - `isDraftOrEditor` (restricts writers to draft documents they created, editors/admins exempt).
  - `isPublishedOrStaff` (allows staff to view drafts, others restricted to published).
- Updated JSDoc documentation at the top of `src/access/policies.ts`.
- Renamed integration test files and test suites under `tests/integration/access/` to align with the `Or` naming pattern (e.g. `isSelfOrAdmin.test.ts`).
- Renamed references in `AGENTS.md`.

#### 3. Test Coverage & Seeding
- Created a dedicated unit test suite for all access control policies in `src/access/__tests__/policies.test.ts`.
- Updated `roles.test.ts` to test pure role checks, explicit admin escalation, and overlapping roles.
- Updated `createUser` test helper and integration test assertions.
- Updated user seeding script (`src/endpoints/seed/users.ts`) to assign arrays of roles.

#### 4. Database Migration
- Created SQL migration script `src/migrations/20260523_010628_migrate_role_to_roles.ts` to safely copy existing users' single role values over to the new `users_roles` join table during migration, and map them back if rolled back.


src/payload.config.ts
```diff
@@ -116,7 +117,7 @@ export default buildConfig({
       path: "/article-recommendations/run",
       method: "post",
       handler: async (req) => {
-        if (!req.user) {
+        if (!req.user || !hasRole(req.user, ADMIN_ROLES)) {
```
tadjh: Can we add a wrapper `isAdmin` so that we don't have to build this everytime?

src/access/policies.ts
```diff
@@ -0,0 +1,68 @@
+import type { Access, Where } from "payload"
+import { hasRole, ADMIN_ROLES, EDITOR_ROLES, STAFF_ROLES } from "./roles"
+
+/**
+ * Access Control Policies
+ *
+ * Naming Conventions:
+ * - Complex state and ownership checks with role exceptions use the format `is[Condition]Unless[ExemptRole]`
+ *   or `is[State]Unless[ExemptRole]` (e.g., `isCreatedByUnlessEditor`, `isPublishedUnlessStaff`, `isDraftUnlessEditor`, `isSelfUnlessAdmin`).
```
tadjh: Thoughts on `Or` instead of `Unless` for policy naming conventions.

`isDraftOrAdmin` > `isDraftUnlessAdmin`

tadjh (changes requested):

tadjh:
Ngl kinda like this one more than the other

src/access/policies.ts
espdesign: Yeah I am open to the change. What do you think reads better in the collection configs?

### 1.  Original
```
export const Articles: CollectionConfig = {
  slug: "articles",
  access: {
    create: writer,
    delete: isCreatedByUnlessEditor,
    read: isPublishedUnlessStaff,
    update: isDraftUnlessEditor,
  },
```

### 2. 
```
export const Articles: CollectionConfig = {
  slug: "articles",
  access: {
    create: writer,
    delete: isEditorOrCreatedBy,
    read: isStaffOrIsPublished,
    update: isEditorOrDraft,
  },
```

### 3. 
```
export const Articles: CollectionConfig = {
  slug: "articles",
  access: {
    create: writer,
    delete: isCreatedByOrEditor,
    read: isPublishedOrStaff,
    update: isDraftOrEditor,
  },
```

Im leaning for number 2, now that you mention `Or`.

espdesign:
> Ngl kinda like this one more than the other

Same, I just did not want to pollute the other PR.

src/payload.config.ts
espdesign: Yeah, I think we reuse the code enough to use a wrapper.

espdesign: However, I also think there may be an argument for not wrapping auth functions as much as possible.

tadjh:
> > Ngl kinda like this one more than the other
> 
> Same, I just did not want to pollute the other PR. 

Do we merge them. What you want to do, the policy version is nice.

src/access/policies.ts
tadjh: I'm thinking #3 since I always favor verb first naming, since you are given the action this function is taking/checking for, then gather the additional context as you keep reading the name.

src/payload.config.ts
tadjh: yeah, in this case I think hasRole and ADMIN_ROLES being the public api for access control depends on two imports for a single logic check. 

I want it to be trivial for devs to reason about and implement access control and for repeated actions like checking if the current user is an admin, we should have a utility function there.

We are using isStaff elsewhere in the codebase and isAdmin should be used similarly.

src/access/policies.ts
espdesign: Im curious what @TedTimbrell thinks.

espdesign:
> > > Ngl kinda like this one more than the other
> > 
> > 
> > Same, I just did not want to pollute the other PR.
> 
> Do we merge them. What you want to do, the policy version is nice.

As soon as we decide on your two points, we can merge.

src/access/policies.ts
tadjh: I think it makes more sense to reason about the primary access, and everything after Unless/Or is an override. Putting the override first is strange to me.

src/payload.config.ts
espdesign: Well we could remove isStaff ;). The abstraction is our const roles categories.

tadjh: I like `isStaff(user)` it's a better consumer API than `hasRole(user, STAFF_ROLES)`

isStaff is consumer code (to be used across the app)
- is single import
- expresses intent in the name

hasRole is library code (to be used internally for access control)
- is highly configurable
- requires importing or creating an array of roles, one of which the user should have


isStaff should use hasRole under the hood.

PR #711, opened 2026-05-22
Add release script and GitHub Actions automation

tadjh:
## Context

Adds a `pnpm release <version>` script and two GitHub Actions to automate the release process end-to-end.

### Local script (`scripts/release.mjs`)

Handles Phase 1 by default — creates a release branch, bumps `package.json`, commits, and opens a PR to `dev`. Phases 2 and 3 remain available as manual overrides via `--phase`:

- `pnpm release 2.3.0` — Phase 1: branch, version bump, PR → dev
- `pnpm release 2.3.0 --phase 2` — Phase 2: PR dev → main (manual override)
- `pnpm release 2.3.0 --phase 3` — Phase 3: tag + release (manual override)

Uses `parseArgs` from `node:util` for argument parsing.

### GitHub Actions

- **`dev-to-main.yml`** — triggered when a version bump lands on `dev`; opens a PR from dev → main and enables auto-merge with a merge commit (not squash) to preserve commit history from dev. Only fires if the actor has `admin` or `maintain` permissions, preventing unauthorized users from triggering the release pipeline.
- **`release.yml`** — triggered when a version bump lands on `main`; creates the git tag and GitHub release automatically.

## Test Plan

- Used this script to create Release v2.2.0
- idk how to test the actions


tadjh:
Should this be an action or is it good the way it is?

espdesign:
> Should this be an action or is it good the way it is?
I think it makes sense as an action imo.

scripts/release.mjs
```diff
@@ -0,0 +1,130 @@
...
+import process from "process"
+import * as readline from "readline"
+import { fileURLToPath } from "url"
+import { blue, gray, green, red, yellow } from "./ansi.mjs"
+
+const __dirname = dirname(fileURLToPath(import.meta.url))
+const root = join(__dirname, "..")
+const pkgPath = join(root, "package.json")
+
+const args = process.argv.slice(2)
```
TedTimbrell: suggestion: consider `import { parseArgs } from 'node:util';`

PR #716, opened 2026-05-24
add back persistance to dockerfile

espdesign:
## Context
Currently our staging previews cannot use the correct url.

> btw had to recover staging database because it ran migrations on it T_T

Also switched to using docker BuildKit secrets as a preventive measure.


dockerfiles/.env.example
```diff
@@ -1,88 +1,69 @@
...
 S3_BUCKET=pragmatic-papers-staging
+# COOLIFY: Build + Runtime + Build Secrets
 S3_ACCESS_KEY_ID=your_access_key_id
 S3_SECRET_ACCESS_KEY=your_secret_access_key
-S3_ENDPOINT=https://s3.amazonaws.com  # or your S3-compatible endpoint
+S3_ENDPOINT=https://s3.amazonaws.com
 
 # ======================
-# DGG Political Action Configuration
+# DGG Political Action
```
tadjh: Why do we have dgg in here still?

dockerfiles/PragmaticPapers.Dockerfile
```diff
@@ -123,33 +91,31 @@ RUN --mount=type=cache,id=nextjs,target=/app/.next/cache \
...
     && addgroup --system --gid 1001 nodejs \
     && adduser --system --uid 1001 nextjs
 
-# Copy the standalone Next.js build
 COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
 COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
 COPY --from=builder --chown=nextjs:nodejs /app/public ./public
 
-# Prepare media directory and set permissions
+# PERSISTENCE FIX: Carry the isolated DATABASE_URI from builder to runner
```
tadjh, line 140: This PR removes a bunch of valuable comments in the Dockerfile, while fixing the persistence.

It helps for future maintainers and for me to understand.

Jarvis is too eager.

tadjh (changes requested):

tadjh:
Let's a few thoughts also, why is the integration test failing?

dockerfiles/PragmaticPapers.Dockerfile
espdesign: Jarvis was just doing what I asked LOL. I can restore them.

dockerfiles/.env.example
espdesign: Good Question.

espdesign:
> Let's a few thoughts also, why is the integration test failing?

https://github.com/digitalgroundgame/pragmatic-papers/pull/717/changes
It just has some bugs/race conditions because each tests share the same database.

tadjh (approved):

PR #726, opened 2026-05-26
Add post hook for migration formatting

TedTimbrell:
## Context

Migrations (at least on my machine) autogenerate with single quotes. While it can just get cleaned up in commit hooks it feels kinda bad and the generated immediate diff is noisy in the migration index.

Any command with 'post' in front of it is called following the success of the command without post.

## Test Plan

Tested locally on my machine.


tadjh (approved):
LGTM

PR #728, opened 2026-05-27
Add Sentry

TedTimbrell:
## Context

Installed using the sentry install wizard (so all autogenerated files)

The DSN urls aren't 'private' (they get bundled into the client code) but I'm making them env variables just to make it slightly more annoying to spam our endpoint and waste our quota. 

## Test Plan
- [x] confirm that source maps are getting uploaded (releases) on sentry
- [x] test an error log (done locally with a built in error)
- [x] Add env variables to prod and staging


src/instrumentation-client.ts
TedTimbrell, line 1: These are pretty much all auto gen files

dockerfiles/PragmaticPapers.Dockerfile
```diff
@@ -93,7 +95,9 @@ ENV NODE_ENV=${NODE_ENV} \
     S3_ENDPOINT=${S3_ENDPOINT} \
     NEXT_PUBLIC_GOOGLE_ANALYTICS_ID=${NEXT_PUBLIC_GOOGLE_ANALYTICS_ID} \
     NEXT_PUBLIC_SERVER_URL=${NEXT_PUBLIC_SERVER_URL} \
-    NEXT_PUBLIC_SUPABASE_URL=${NEXT_PUBLIC_SUPABASE_URL}
+    NEXT_PUBLIC_SUPABASE_URL=${NEXT_PUBLIC_SUPABASE_URL} \
+    NEXT_PUBLIC_SENTRY_DSN=${NEXT_PUBLIC_SENTRY_DSN} \
+    NEXT_PUBLIC_SENTRY_ENVIRONMENT=${NEXT_PUBLIC_SENTRY_ENVIRONMENT}
```
TedTimbrell, line 100: environment is set so we can differentiate our staging/preview builds from prod

next.config.ts
```diff
@@ -131,4 +132,40 @@ const nextConfig: NextConfig = {
...
+
+  // For all available options, see:
+  // https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/
+
+  // Upload a larger set of source maps for prettier stack traces (increases build time)
+  widenClientFileUpload: true,
+
+  // Route browser requests to Sentry through a Next.js rewrite to circumvent ad-blockers.
+  // This can increase your server load as well as your hosting bill.
+  // Note: Check that the configured route will not match with your Next.js middleware, otherwise reporting of client-
```
tadjh: Is our proxy.ts (formerly known as middleware) matchers configured for sentry a according to this comment?

TedTimbrell: It is not. Good catch

src/instrumentation-client.ts
```diff
@@ -0,0 +1,33 @@
...
+  // Enable logs to be sent to Sentry
+  enableLogs: true,
+
+  // Define how likely Replay events are sampled.
+  // This sets the sample rate to be 10%. You may want this to be 100% while
+  // in development and sample at a lower rate in production
+  replaysSessionSampleRate: 0.1,
+
+  // Define how likely Replay events are sampled when an error occurs.
+  replaysOnErrorSampleRate: 1.0,
```
TedTimbrell: We only get like 50 of these a month. I imagine we'll tune this down or disable it.

src/instrumentation-client.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as Sentry from "@sentry/nextjs"
+
+Sentry.init({
+  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+
+  environment: process.env.NEXT_PUBLIC_SENTRY_ENVIRONMENT ?? process.env.NODE_ENV,
+
+  tracesSampleRate: 0.1,
+
+  // Enable logs to be sent to Sentry
```
TedTimbrell:
```suggestion
```

src/instrumentation-client.ts
```diff
@@ -0,0 +1,22 @@
...
+  enableLogs: true,
+
+  // Enable sending user PII (Personally Identifiable Information)
```
TedTimbrell:
```suggestion
```

next.config.ts
```diff
@@ -131,4 +132,40 @@ const nextConfig: NextConfig = {
...
+  // side errors will fail.
+  tunnelRoute: "/monitoring",
+
+  webpack: {
+    // Enables automatic instrumentation of Vercel Cron Monitors. (Does not yet work with App Router route handlers.)
+    // See the following for more information:
+    // https://docs.sentry.io/product/crons/
+    // https://vercel.com/docs/cron-jobs
+    automaticVercelMonitors: true,
+
```
TedTimbrell:
```suggestion
```

sentry.edge.config.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as Sentry from "@sentry/nextjs"
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
+
+  environment: process.env.NEXT_PUBLIC_SENTRY_ENVIRONMENT ?? process.env.NODE_ENV,
+
+  tracesSampleRate: 0.1,
+
+  // Enable logs to be sent to Sentry
```
TedTimbrell:
```suggestion
```

sentry.edge.config.ts
```diff
@@ -0,0 +1,21 @@
...
+  enableLogs: true,
+
+  // Enable sending user PII (Personally Identifiable Information)
```
TedTimbrell:
```suggestion
```

sentry.server.config.ts
```diff
@@ -0,0 +1,20 @@
...
+import * as Sentry from "@sentry/nextjs"
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
+
+  environment: process.env.NEXT_PUBLIC_SENTRY_ENVIRONMENT ?? process.env.NODE_ENV,
+
+  tracesSampleRate: 0.1,
+
+  // Enable logs to be sent to Sentry
```
TedTimbrell:
```suggestion
```

sentry.server.config.ts
```diff
@@ -0,0 +1,20 @@
...
+  enableLogs: true,
+
+  // Enable sending user PII (Personally Identifiable Information)
```
TedTimbrell:
```suggestion
```

dockerfiles/PragmaticPapers.Dockerfile
tadjh: Are there ever times where Sentry Environment will be different from BUILD_ENV or NODE_ENV. Really asking if there's a reason we can't use those instead?

TedTimbrell: I'm just trying to avoid checking the URL into github but I think I'm misunderstanding.

tadjh (dismissed):

tadjh:
LGTM

tadjh (commented):
This looks good just a few things:

1. sendDefaultPii: true — needs intentional sign-off

This is set in all three init files. It causes Sentry to capture IP addresses, cookies, HTTP headers, and usernames. This has GDPR/CCPA implications. If the site has EU users, you need either:

User consent before Sentry initializes, or
sendDefaultPii: false (the safer default)
This should be an explicit, documented team decision — not just wizard output left in.

2. tracesSampleRate: 0.1 — verify this is intentional

10% trace sampling is reasonable for a low-traffic site, but this controls both performance monitoring and error rate context. Consider whether you want tracesSampleRate: 1.0 initially to get full visibility, or confirm 10% is acceptable. Errors are always captured regardless; this only affects performance spans.

tadjh (approved):

PR #729, opened 2026-05-29
Prevent staging from getting search indexed

TedTimbrell:
## Context

Noticed that a couple of staging pages were getting crawled/indexed. They were never ranked highly but it's worth not polluting our SEO search space.

Adds a `no-index, nofollow` robots header. Counterintuitively, blocking crawling might cause google to keep the page indexed.


tadjh (approved):
LGTM

PR #768, opened 2026-06-04
Add InteractiveMap block for displaying choropleth maps

### LINKED ISSUE #764: Create Interactive Map Block
Request from Emily: Missouri interactive map for article. Generalized into a reusable Interactive Map block for the journal.

## Original requirements (from Emily)

- Use [D3](https://d3js.org/getting-started#d3-in-react) for map rendering of SVG data
- Render a map within an article
- Render multiple maps side by side in one widget
- Give maps titles
- State districts are hoverable to reveal a tooltip (data structure TBD — defined below)
- Projection: ESRI:102003 Albers equal area; fall back to NAD83(2011) state projections when Albers is unavailable
- Red and Blue shaded regions; color support
- Watermark of the sources, as required by the sources of data

Example HTML provided by Emily: [mo_districts_demo.html](https://github.com/user-attachments/files/28580655/mo_districts_demo.html). The demo inlines pre-projected SVG (Albers baked in at data-prep time) with `data-district` / `data-margin` attributes on each `<path>`, a step color scale on margin, and a shared tooltip. No D3 at runtime in the demo — we need to support that flow AND a D3-driven flow that projects from raw GeoJSON.

Example data sources cited in comments:
- Right-side map: https://redistrictingdatahub.org/terms-and-conditions/
- Left-side map: https://cdmaps.polisci.ucla.edu/

---

### Tech choice

Use targeted D3 packages — **`d3-geo`, `d3-scale`, `d3-scale-chromatic`, `topojson-client`** (~25 KB gzipped combined). React owns rendering; D3 owns the math (projections + scales). We do not pull in `d3-selection` or other DOM-manipulation parts. Gestures (hover / focus / touch) are written with React event handlers.

### Block placement

Register the block in BOTH:
- Lexical `BlocksFeature` in `src/collections/Articles/index.ts` — writers insert it inline within article body.
- Pages `layout` blocks in `src/collections/Pages/index.ts` + `src/blocks/RenderBlocks.tsx` — for full-bleed standalone dashboards.

### Geometry "source adapter" model

Block field `geometrySource` is a discriminated union with three v1 adapters:

| Adapter | Writer experience | Implementation |
|---|---|---|
| `upload` | Upload `.geojson` / `.topojson` via Media | Extend Media mimetype allowlist; D3 projects at runtime |
| `inlineSvg` | Paste pre-projected SVG (Emily's current workflow) | Sanitize + inline; expect `data-region` attributes on paths |
| `remoteApi` | Pick a known source (Census TIGERweb, UCLA cdmaps) + params in the admin. Server-side fetch + persist GeoJSON to Media on first save. From then on it behaves like `upload`. | Small registry of `RemoteSource` configs; Payload `beforeChange` hook does the fetch and rewrites the block to point at the new Media doc. |

The `remoteApi` adapter is the key win — our library *builds itself* as writers query external sources, no manual upload chore, and the fetched data is reusable across articles.

Adapters all return the same shape to the renderer: `{ features: Array<{ id, label, path: string }>, viewBox, attribution }`. Adding a fourth source (ArcGIS REST, MapBox vector tiles, …) is a single PR.

#### v1 remote sources

- **U.S. Census TIGERweb REST** — states, counties, congressional districts. Public domain, no key required. Returns ArcGIS JSON; normalize via `arcgis-to-geojson-utils`.
- **UCLA cdmaps** — historic Congressional districts. Downloadable shapefiles, not a queryable API. Treat as a hard-coded list of known file URLs the writer picks from; fetch + convert with `shpjs` on first use. Honor their attribution requirements.

Source shape:

```ts
type RemoteSource = {
  id: string;
  name: string;
  attribution: { name: string; url: string };
  params: ReadonlyArray<{
    key: string; label: string;
    options: Array<{ value: string; label: string }>;
  }>;
  buildUrl: (params: Record<string, string>) => string;
  toGeoJson?: (raw: unknown) => GeoJSON.FeatureCollection;
};
```

Out of scope for v1 (registry supports them later): Redistricting Data Hub (requires login), MapBox, private ArcGIS layers.

### Data binding

Default to a Payload repeater on each map: `regions: [{ regionId: string, label?: string, value?: number, color?: string, tooltipRows?: Array<{ label, value }> }]`. `regionId` joins to the feature's `id` (or a writer-specified property key for GeoJSON, or the `data-region` attribute for inline SVG). Editable inline in admin; live preview works; no separate file required.

Future option (defer): allow a Media-uploaded CSV/JSON sidecar for very large datasets.

### Color scale

Block fields:
- `colorScale.type`: `diverging` (R/D margin) | `sequential` | `categorical`
- `colorScale.breakpoints`: numeric stops (default to demo's `[1, 5, 15]` for diverging)
- `colorScale.palette`: named preset (`red-blue` matching the demo, `viridis`, `single-hue-blue`, …) OR custom hex array
- `colorScale.neutralFill` for regions with no data

Implement with `d3-scale.scaleThreshold` (diverging / sequential) or `scaleOrdinal` (categorical). Ship Emily's exact diverging palette as the default so her article renders identically out of the box:

```
abs<1   → R+ #cd897f / D+ #9499b2
abs<5   → R+ #fd8997 / D+ #89aefd
abs<15  → R+ #fd5864 / D+ #587ac9
else    → R+ #b7212c / D+ #22428c
```

### Tooltip schema

Explicit (this was "undecided" in the original):
- `title` — region `label`, falls back to `regionId`
- `subtitle` — formatted from `value` via the active color scale (e.g. `R+2.4` for diverging margins, raw number otherwise)
- `rows` — optional `tooltipRows[]` for extra `label: value` lines

One tooltip DOM node shared across all maps in the widget (matches the demo).

### Widget structure (multi-map)

The block is a **widget** containing 1..N maps:
- `widgetTitle`, `widgetDescription` (rich text)
- `maps: [{ title, geometrySource, regions[] }]` (repeater)
- `layout`: `row` | `grid` (responsive — wraps on narrow viewports)
- `sources: [{ name, url }]` — rendered as a watermark/attribution footer (the "watermark of the sources" requirement). Auto-populated from each map's `geometrySource.attribution` so the writer doesn't have to type it.

### Projection

When D3 projects from GeoJSON (`upload` and `remoteApi` post-persist):
- `projection`: `albers-usa` | `albers-equal-area` (rotated to the state's centroid automatically) | `mercator` | `identity` (for already-projected coordinates)
- Default to `albers-equal-area` for state-level data to honor Emily's ESRI:102003 preference. Fit-extent to the viewBox automatically.

Bypassed entirely by `inlineSvg` (writer has already projected).

### Accessibility & gestures

- Regions are keyboard-focusable (`tabindex=0`, focus ring); Enter/Space pins the tooltip.
- Touch: tap to pin tooltip; tap elsewhere dismisses.
- Each region gets a `<title>` child for screen readers (`{label}: {formatted value}`).
- Block-level `aria-label` summarizing the widget.

---

## Implementation outline

Mirror existing block conventions (see `src/blocks/MediaBlock/`, `src/blocks/Code/`, `src/blocks/SocialEmbed/`):

1. `src/blocks/InteractiveMap/config.ts` — Payload block config (slug `interactiveMap`, fields above).
2. `src/blocks/InteractiveMap/Component.tsx` — async server component. Resolves geometry via the adapter, computes the color scale server-side, renders SVG markup, passes interaction props to a client child.
3. `src/blocks/InteractiveMap/Component.client.tsx` — `"use client"`. Owns tooltip state, pointer / keyboard / touch handlers, hover-overlay path (mirrors the demo's `mo-hover-outline` pattern).
4. `src/blocks/InteractiveMap/adapters/{upload,inlineSvg,remoteApi}.ts` — each exports `resolve(source): Promise<ResolvedGeometry>`.
5. `src/blocks/InteractiveMap/remoteSources/{censusTigerweb,uclaCdmaps}.ts` + `index.ts` — registry of `RemoteSource` configs.
6. Payload `beforeChange` hook on the block — when `geometrySource.type === 'remoteApi'` and there's no persisted Media reference yet, fetch the URL server-side, convert if needed, save as a Media doc (filename derived from source id + params), rewrite the block to point at it. Subsequent renders go through the `upload` code path.
7. `src/blocks/InteractiveMap/projection.ts` — wrapper around `d3-geo` for choosing/fitting a projection.
8. `src/blocks/InteractiveMap/colorScale.ts` — wrapper around `d3-scale` exposing `(value) => color` and `(value) => formattedLabel`.
9. Register in Lexical `BlocksFeature()` in `src/collections/Articles/index.ts` (~lines 122–127).
10. Register in Pages `layout` field in `src/collections/Pages/index.ts` and the switch in `src/blocks/RenderBlocks.tsx`.
11. Register in RichText converters in `src/components/RichText/index.tsx` (~lines 68–112).
12. Extend Media collection to accept `application/geo+json`, `application/json`, `image/svg+xml` (`src/collections/Media/index.ts:118–192`).
13. Run `pnpm payload generate:types`.

---

## Security / safety

- **SVG sanitization** (`inlineSvg`): use `DOMPurify` server-side or a tight allowlist — allow `<g>`, `<path>`, `data-*` only; strip `<script>`, event handlers, external refs.
- **`remoteApi` fetch**: server-side only, response size capped, ~15s timeout, outbound URLs restricted to the registered source allowlist (no arbitrary URLs from the admin form). The registry is the security boundary.
- **Dedupe**: tag fetched Media docs with `source.id` + `source.params` so we never refetch the same Census-2024 state boundary twice, and so we can later surface a "library" UI of everything we've collected.

---

## v1 acceptance criteria

- [ ] Block available as a Lexical block inside Article body content.
- [ ] Block available as a Pages layout block.
- [ ] `inlineSvg` adapter renders Emily's `mo_districts_demo.html` SVG_119 / SVG_120 strings identically (hover overlay, tooltip content, palette).
- [ ] `upload` adapter accepts a GeoJSON via Media, D3 projects with `albers-equal-area` defaulted for state-level data.
- [ ] `remoteApi` adapter fetches U.S. Census TIGERweb (state, county, CD) on first save, persists GeoJSON to Media, dedupes on `source.id + params`.
- [ ] Multi-map "widget" supports N maps in a row with a shared color scale, shared tooltip, per-map titles.
- [ ] Attribution footer auto-populated from each `geometrySource.attribution`; writer can add overrides.
- [ ] Keyboard-only navigation: Tab through regions, Enter/Space pins tooltip, Esc dismisses.
- [ ] Touch: tap pins tooltip, tap elsewhere dismisses.
- [ ] Snapshot tests for the renderer; integration test for `upload` with a fixture GeoJSON and for `remoteApi` with `msw`-mocked Census responses.

## Out of scope for v1 (track separately)

- Point markers / bubble maps
- Time-series / animated playback
- CSV/JSON sidecar data uploads
- Redistricting Data Hub source (auth required)
- MapBox / private ArcGIS layers
- Admin "library" UI listing all auto-fetched Media docs (can come once we have enough of them)

### LINKED ISSUE #765: Interactive Map Phase A — inlineSvg adapter + Lexical block
Phase A of #764 — first deliverable so Emily can upload her existing pre-projected SVGs and ship her Missouri article.

## Scope

- New \`InteractiveMap\` block at \`src/blocks/InteractiveMap/\`
- New \`MapAssets\` Payload collection (\`map-assets\` slug) — SVG file upload; a \`beforeValidate\` hook captures \`svgContent\` at upload time so blocks read it without a runtime fetch
- \`inlineSvg\` adapter: server fetches \`svgContent\` from the uploaded asset, sanitizes with a strict allowlist (svg/g/path/title/desc; \`data-*\` and \`aria-*\` whitelisted; \`viewBox\` case preserved), and parses paths with \`htmlparser2\`
- \`dataAttribute\` field on each map — names the \`data-*\` attribute on SVG paths that holds the numeric value for the color scale (e.g. \`data-margin\`); values read directly from the SVG, no manual entry needed
- \`overrides[]\` field on each map — per-region \`label\`, \`value\`, and \`color\` overrides for cases not covered by \`dataAttribute\`
- \`colorBias\` field — multiplier that scales color intensity independently of the underlying margin values
- \`divergingRedBlue\` + \`perRegion\` color scales — Emily's MO palette baked in as the default
- Shared tooltip with R+/D+ formatting, black hover-outline overlay; **multiple pinned tooltips** — click to pin a region's tooltip, click another to stack a second; Escape clears all
- Keyboard-focusable regions, \`<title>\` labels for screen readers
- Multi-map "widget" layout (\`row\` / \`grid\`) with shared color scale + shared tooltip
- Source attribution footer
- Registered in the Lexical editor on Articles (NOT yet on Pages — that comes in Phase B)
- Unit tests for \`colorScale\`, \`sanitize\`, \`parseInlineSvg\`, and the \`inlineSvg\` adapter
- Snapshot test for the server component + interaction tests for the client component
- Playwright E2E tests for tooltip hover and click-to-pin

## Out of scope (later phases)

- \`upload\` adapter (GeoJSON → D3 projection) → Phase B (#766)
- \`remoteApi\` adapter (Census TIGERweb + UCLA cdmaps) → Phase C (#767)
- Pages layout-block registration → Phase B
- Media collection mimetype allowlist for GeoJSON → Phase B

## Acceptance

- [x] Block available in the Lexical editor on Articles
- [x] SVG upload to \`map-assets\` collection; \`svgContent\` captured at upload time
- [x] Inline SVG paths colored by region data via \`dataAttribute\`, R+/D+ tooltips, hover overlay
- [x] Side-by-side multi-map widget with shared color scale + tooltip; multiple pinned tooltips supported
- [x] Keyboard nav (Tab/Enter/Esc) + screen-reader \`<title>\`
- [x] Source attribution footer
- [x] Unit + snapshot + E2E tests pass; \`pnpm lint:ci\` + \`pnpm check-types\` clean

Branch: \`claude/election-maps-d3js-eWcO6\`

Parent: #764

tadjh:
## Context

This PR adds a new `InteractiveMap` block that enables editors to create interactive choropleth maps with hover tooltips and click-to-pin functionality. The block supports:

- **Multiple map layouts**: Side-by-side (row) or grid layout with responsive wrapping
- **Color scales**: Diverging red/blue scale for election margins (R+/D+) or per-region custom colors
- **Interactive features**: Hover tooltips, keyboard navigation, click-to-pin with escape-to-close
- **SVG flexibility**: Accepts pre-projected SVG with configurable region attributes
- **Attribution**: Optional sources/attribution footer

The implementation includes:
- Server-side SVG sanitization with a strict allowlist to prevent XSS
- Client-side interactivity with pointer and keyboard event handling
- Color scale utilities for diverging margins and per-region overrides
- Comprehensive unit tests for color scales, SVG sanitization, and component rendering
- Full TypeScript support with generated Payload CMS types

## Test Plan

Added comprehensive unit tests covering:
- **Color scale logic** (`colorScale.test.ts`): Diverging color picker, margin formatting, and color resolution
- **SVG sanitization** (`sanitize.test.ts`): Script stripping, event handler removal, data attribute preservation
- **Map resolution** (`inlineSvg.test.ts`): SVG sanitization, region color/value computation, attribute defaults
- **Component rendering** (`Component.snapshot.test.tsx`): Server-side rendering with snapshot verification

All tests pass and cover the core functionality. The block integrates with existing Payload CMS infrastructure and RichText component system.

https://claude.ai/code/session_01K4BTTqwREQhyFjXcdevRVU

https://pr-768.pragmaticpapers.com/articles/missouris-shifting-margins-the-119th-and-120th-congressional-maps


tadjh:
https://coolors.co/000000-ffffff-ff401a-006ee3-cc0026

EmilyCapper:
- [ ] map display box is confined tightly around the map geometry, such that hover-outlines clip with the box
- [ ] dark mode margin colors are harsh on the eyes. will require some experimentation.
team crimson/red (these might not work well but it's worth a shot) (order: tilt, lean, likely, safe)
#b08090
#f075a0
#cc2a5a
#8a1038
team cobalt/blue
#8aa5c0
#70b8f5
#3080d0
#1548a0
color_palette

- [ ] style: for dark mode, try dark grey district borders with white outline on hover, instead of white borders and white outline

tadjh:
Option 1
- `data-margin` is hardcoded to always be `R+D-`
Option 2
- inside of payload you select `data-margin` and choose `R+D-`

tadjh:
updated the colors based on dgg branding guide and also added a color bias (0.5) feature so the distinction between deep red and deeeeeper red is more obvious 
image

EmilyCapper:
> updated the colors based on dgg branding guide and also added a color bias (0.5) feature so the distinction between deep red and deeeeeper red is more obvious image

I disagree with the color brackets change.
- [ ] I think we should revert the third red to a margin between 5 and 15 (and so on), then add an optional R+40 tier.

The way that the four shades each of red and blue are used in the online election data community ("safe" as 15+ margins, "likely" as 5-15, "lean" as 1-5, and "tilt" as 0-1) will make our figures harder to interpret due to starting a different standard, and might suggest to some that certain districts are way more flippable than they actually are (R+20 and harder districts are not generally flippable). I'm really on the fence about adding a fifth R+40 and above tier color. On one hand, it can be nice as a style when you're looking at the breadth of district leanings in a state, and on the other, the five, seven, and eight-color standard grading systems vary how fine your categories are within the ~95-99% of district flips that actually occur, which is nice for election forecasting. I think it would be okay to have the R+40 tier be optional, and I would switch it on for the Missouri article because the purpose fits. And speaking of feature changes...
- [ ] We should add a color scale bar like FiveThirtyEight's (details below)
image

They have fewer margin color tiers in their redistricting tracker, and I would leave off the margin bracket names, but using it as a visual aid for a concept: Having eight, fairly thin color rectangles lined up directly next to each other, with thin dark grey/white (light/dark mode) lines coming down from the boundary to indicate the boundary margin (and also at the very ends of the bar, slightly longer and unlabeled), and the R+/D+ markers represent (sometimes -40,) -15, -5, -1, 0 and so on. Because it is an exact zero boundary in this case, the R+/D+0 label should be substituted with EVEN. Though, I'm not sure which is better: placing it above the map or between the map and watermark. If the ticks don't work, we can switch it to >=margins below the middle of the rectangle and pull the rectangles apart.

tadjh:
> ## Snapshot updates in this PR
> interactive-map-tooltip

@EmilyCapper 

- Breakpoints defaulting to [1,5,15,30]
- Neutral is +1 or -1
- Added bias to shift the color breakpoints with 0 and 100 being pinned. This is to demo how the map looks with full color steps. Bias scales legend as well. Bias value shows in the tooltip of the `info` icon when hovered.
- Added color invert to test blue colors since this data set is R+ heavy

The final article version will be correct, these are just custom configurations I made on our test articles in the repo to showcase all colors.

https://pr-768.pragmaticpapers.com/articles/missouris-shifting-margins-the-119th-and-120th-congressional-maps

EmilyCapper:
> > ## Snapshot updates in this PR
> > interactive-map-tooltip
> 
> @EmilyCapper
> 
> * Breakpoints defaulting to [1,5,15,30]
> * Neutral is +1 or -1
> * Added bias to shift the color breakpoints with 0 and 100 being pinned. This is to demo how the map looks with full color steps. Bias scales legend as well. Bias value shows in the tooltip of the `info` icon when hovered.
> * Added color invert to test blue colors since this data set is R+ heavy
> 
> The final article version will be correct, these are just custom configurations I made on our test articles in the repo to showcase all colors.
> 
> https://pr-768.pragmaticpapers.com/articles/missouris-shifting-margins-the-119th-and-120th-congressional-maps

Yea this looks brilliant. Having +/-30 as the threshold makes sense. I would guess if you add 0 as a breakpoint, it will generate an even number of blocks? Also, I overlooked before that having an exact zero behave as EVEN would cause issues in this context. Since the zero is "split," I suppose, R+0 and D+0 are fine. A consequence of an even number of margin brackets.

tadjh:
> > > ## Snapshot updates in this PR
> > > interactive-map-tooltip
> > 
> > @EmilyCapper
> > 
> > * Breakpoints defaulting to [1,5,15,30]
> > * Neutral is +1 or -1
> > * Added bias to shift the color breakpoints with 0 and 100 being pinned. This is to demo how the map looks with full color steps. Bias scales legend as well. Bias value shows in the tooltip of the `info` icon when hovered.
> > * Added color invert to test blue colors since this data set is R+ heavy
> > 
> > The final article version will be correct, these are just custom configurations I made on our test articles in the repo to showcase all colors.
> > 
> > https://pr-768.pragmaticpapers.com/articles/missouris-shifting-margins-the-119th-and-120th-congressional-maps
> 
> Yea this looks brilliant. Having +/-30 as the threshold makes sense. I would guess if you add 0 as a breakpoint, it will generate an even number of blocks? Also, I overlooked before that having an exact zero behave as EVEN would cause issues in this context. Since the zero is "split," I suppose, R+0 and D+0 are fine. A consequence of an even number of margin brackets. 

Data between + or - 1 point is considered neutral

PR #770, opened 2026-06-05
Move mode selector up to nav

TedTimbrell:
## Context

While I generally love dark mode for most things I am a convert to longform only being readable in light mode. 

* Moves the theme selector to the header.
* Changes the right hand sheet icon to a gear ⚙️  to indicate settings and personalisation
* Adds a full width mode for theme selection on mobile
* Fixes a bug where hovering over the theme didn't change the cursor i.e. it didn't seem like an actionable icon. 

Screenshot 2026-06-04 at 9 41 47 PM
Screenshot 2026-06-04 at 9 41 41 PM
Screenshot 2026-06-04 at 9 45 51 PM


src/Header/Component.tsx
```diff
@@ -72,12 +73,15 @@ export async function Header(): Promise<React.JSX.Element> {
...
+                <ModeToggle />
+              </div>
               <Sheet>
                 <SheetTrigger
                   render={
                     <Button variant="ghost" size="icon" className="lg:hidden">
-                      <User className="size-6" />
-                      <span className="sr-only">Account</span>
+                      <Settings className="size-6" />
+                      <span className="sr-only">Settings</span>
```
tadjh: Notice nytimes and bbc use a User Icon instead of cog wheel. It was implemented it match them, but we could lean more tech/saas with the cog wheel

src/Footer/Component.tsx
```diff
@@ -17,7 +16,6 @@ export async function Footer(): Promise<React.ReactElement> {
         </a>
         <div className="flex flex-row items-center gap-2">
           <SocialLinks parentId={id} socials={socials} aria-label="Footer Social Links" />
-          <ModeToggle />
```
tadjh, line 26: I'm thinking we keep the footer one and add the header one as well?

src/Header/Component.tsx
espdesign: I'm not a huge fan of "settings" for this menu.

tadjh:
@TedTimbrell should we just ignore system default and instead default to Light mode? and leave dark mode for those who prefer it at the bottom?

src/Header/Component.tsx
TedTimbrell: Yeah sounds good I'll revert this specific one. I'm gonna change the screen reader text to "Account and settings" though

TedTimbrell:
> @TedTimbrell should we just ignore system default and instead default to Light mode? and leave dark mode for those who prefer it at the bottom?

I mean if you're willing we can. I imagine we'll get people yelling at us in chat initially lmao. 


Also on the desktop header. I think the toggle looks better to the right of our existing buttons? @tadjh I know you objected.

I still think it would be kinda odd to put it on the left next to the hamburger menu.

Screenshot 2026-06-13 at 4 45 26 PM
Screenshot 2026-06-13 at 4 45 41 PM

PR #772, opened 2026-06-05
Trigger lightbox only on media selection

TedTimbrell:

### DIFF
--- a/src/blocks/MediaBlock/LightboxMediaBlock.tsx
+++ b/src/blocks/MediaBlock/LightboxMediaBlock.tsx
@@ -22,15 +22,16 @@ export const LightboxMediaBlock: React.FC<LightboxMediaBlockProps> = ({
 
   return (
     <Dialog>
-      <DialogTrigger className={cn("w-full cursor-pointer", containerClassName)}>
-        <MediaBlock
-          media={media}
-          enableGutter={false}
-          className={className}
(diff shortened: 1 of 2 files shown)


src/blocks/MediaBlock/LightboxMediaBlock.tsx
```diff
@@ -22,15 +22,18 @@ export const LightboxMediaBlock: React.FC<LightboxMediaBlockProps> = ({
...
+        media={media}
+        enableGutter={false}
+        className={cn("w-full", containerClassName, className)}
+        disableInnerContainer
+        mediaWrapper={(mediaNode) => (
+          <DialogTrigger className="block w-full cursor-pointer [&>*]:m-0">
+            {mediaNode}
+          </DialogTrigger>
+        )}
+        {...props}
```
tadjh, line 35: lets use composition here instead of a mediaWrapper render function, DialogTrigger and Media can be nested directly

TedTimbrell: Sorry I don't quite follow. 

Are you saying to pass a node instead of a function? Or are you saying we should pass Media and DialogTrigger as children?

tadjh: Pass as JSX children instead of a props

PR #802, opened 2026-06-16
Upgrade to PG 17 for local and tests

### LINKED ISSUE #795: Postgres is two major versions behind
The dev database runs on Postgres 15 while the current stable is 17, creating a gap between local and production environments.

TedTimbrell:
## Context
Closes #795 Prod uses pg 17

* Upgrades to 17-alpine
* Separates db-nuke and db-fresh. The former hard-deletes volumes while the latter goes through the payload destruction and creation flow. The latter is what we'd want to normally use.

Use `pnpm dev:db-nuke` on local for the upgrade.

After the first nuke, use `pnpm dev-db:fresh` going forward


package.json
tadjh, line 1: the `&& pnpm dev:db-down` made it so when you finish migrate:fresh it brings the db down. Might be worth keeping for dev:fresh

tadjh (changes requested):

package.json
```diff
@@ -10,7 +10,8 @@
     "postbuild": "next-sitemap --config next-sitemap.config.cjs",
     "dev": "pnpm dev:db && pnpm dev:next && pnpm dev:db-down",
     "dev:db": "docker compose -p pragmatic-papers -f docker-compose.yml up -d",
-    "dev:db-nuke": "pnpm dev:db && echo y | pnpm payload migrate:fresh && pnpm dev:db-down",
+    "dev:db-nuke": "docker compose -p pragmatic-papers -f docker-compose.yml down -v",
+    "dev:db-fresh": "pnpm dev:db && echo y | pnpm payload migrate:fresh",
```
tadjh:
```suggestion
    "dev:db-fresh": "pnpm dev:db && echo y | pnpm payload migrate:fresh && pnpm dev:db-down",
```

tadjh (approved):

PR #833, opened 2026-07-05
Dockerized local E2E screenshot baselines + pin CI to same image

tadjh:
## Context

Every time a screenshot test is added or intentionally changed, the workflow today is: push, wait for CI, let it auto-commit a baseline (or trigger the "needs screenshots" label workflow). That's slow to iterate on and adds noisy bot commits/comments to the PR.

This PR adds a way to generate Playwright screenshot baselines **locally** with parity to CI, so routine baseline updates don't need the push-wait-auto-commit loop:

- **`docker-compose.e2e.yml`** + **`scripts/test-e2e-docker.sh`** (`pnpm test:e2e:update-snapshots`) run the E2E suite inside `mcr.microsoft.com/playwright:v1.60.0-noble` — the official Playwright image pinned to the exact `@playwright/test` version resolved in `pnpm-lock.yaml`. Postgres and the Playwright image run as sibling compose services (no host networking, no Docker socket mounts — works on Docker Desktop for Mac/Windows too, and matches how Playwright's CI docs structure containerized runs). Inside the Playwright container it installs deps (node_modules in a named volume so host-platform native binaries aren't clobbered), migrates/seeds, and runs `node scripts/test-e2e.mjs` with `E2E_PROD_SERVER=true` — the same production-build flow CI uses. `GH_FONT_READ` is passed through (CI renders with the private fonts package installed, so fallback-font baselines would never match) with a warning when unset.
- **`.github/workflows/playwright.yml`** and **`.github/workflows/update-snapshots.yml`** now run their test jobs inside that same pinned container image, instead of installing Chromium fresh on bare `ubuntu-latest` via `playwright install --with-deps`. That install was never fully reproducible (GitHub can silently update the runner image/fonts/system libs over time); pinning both CI and local tooling to the identical image makes them deterministically identical instead of "hopefully close." Removed the now-redundant "Install/cache Playwright browsers" steps since the image already bundles Chromium.
- `DATABASE_URI` in both workflows now points at the `postgres` service hostname instead of `localhost`, since the test job runs inside a `container:` and service containers are reachable by their service-id alias on the shared Docker network, not the runner's loopback interface.
- **`scripts/check-playwright-image-pin.mjs`** guards against the pin going stale: it fails if the image tag in `docker-compose.e2e.yml` or either Playwright workflow drifts from the `@playwright/test` version pnpm actually resolved in `pnpm-lock.yaml`. It runs in its own advisory workflow (`.github/workflows/playwright-pin-check.yml`) — triggered on `dev`/`main` pushes touching the relevant files plus a weekly schedule — rather than as part of `ci.yml`'s Static checks, since bumping the Playwright pin is a deliberate chore and shouldn't gate every feature branch's required checks if drift ever landed on trunk.
- **`tests/e2e/README.md`** and **`AGENTS.md`** updated to describe the new supported local (Dockerized) path, while keeping the warning against generating baselines on a bare host OS.

## Test Plan

- `pnpm lint:ci` and `pnpm check-types` pass.
- Validated all workflow/compose YAML parses correctly and the shell script passes `bash -n`.
- Verified `scripts/check-playwright-image-pin.mjs` passes on the current pin and correctly fails (with a clear error message) when a tag is deliberately mismatched.
- **Could not exercise Docker locally in this sandbox** (no Docker daemon available in the execution environment used to write this PR) — I was unable to actually run `pnpm test:e2e:update-snapshots` or byte-diff a locally-generated baseline against an existing CI baseline before opening this PR. Please have CI actually run against this branch to confirm the `container:`/`services:` networking works as expected (GitHub Actions has sharp edges here), and if possible have someone with local Docker try `pnpm test:e2e:update-snapshots` (with `GH_FONT_READ` exported) against an existing screenshot test and compare the result to committed baselines before relying on this for new baselines.


tadjh:
Not sure about this one. Need to think about the implications, how to update to latest, and what happens when baselines drift because of it.

PR #860, opened 2026-07-16
Claude's attempt at remove the n+1 query

### LINKED ISSUE #861: Fix N+1 Query on Articles
Sentry Issue: [PRAGMATIC-PAPERS-3](https://digital-ground-game.sentry.io/issues/7508515736/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Offending Spans** | db - pg-pool.connect

TedTimbrell:


tadjh:
## Review: is this the idiomatic way to dedupe the N+1?

This PR is really two different fixes, and they land differently on the "idiomatic" scale.

### 1. `populatedVolume` hook → `join` field ✅ (the strong part)

Replacing the per-article `populateVolume` afterRead hook (one `find` against `volumes` *per article* — the actual N+1) with a `join` field is exactly what Payload 3's `join` type exists for. Payload populates joins natively and batched, and it lets the PR delete the hook, the `populatedVolume` virtual group, and the `skipPopulateVolume` seed plumbing. This is the change I'd keep with the least hesitation.

**One regression to verify:** the old hook queried volumes *unconditionally*, so `populatedVolume` was always present. A `join` only populates when the query reaches it at sufficient depth. `buildArticleJsonLd` now does:

```ts
const volumeDoc = article.volume?.docs?.find((d): d is Volume => typeof d !== "number")
```

If the query feeding `structuredData` fetches the article at a depth where `volume.docs` come back as numeric IDs (or the join is excluded by a `select`), `isPartOf` silently drops out of the JSON-LD. Worth confirming that path actually populates the join.

### 2. `populateAuthors` / `populateNarrator` → custom `getRequestLoader` ⚠️ (debatable)

**Why a custom loader is needed at all:** authors/narrator are *forward* relationships, so Payload's native depth population would already batch them through its per-request DataLoader — except `users` is access-locked, so they need `overrideAccess: true`. That's a legitimate reason to hand-populate, and the fix correctly targets the Payload layer (this is a loop inside one `payload.find`, so Next's `React.cache`/`unstable_cache` wouldn't touch it).

But the specific implementation has some sharp edges:

- **It reinvents `dataloader`.** Payload already depends on Facebook's `dataloader` and exposes a live instance as `req.payloadDataLoader`. `getRequestLoader` hand-rolls the same queueMicrotask "batch-within-a-tick" pattern from scratch, untested.
- **Batching is best-effort, not guaranteed-single-query.** Coalescing only works if every article's `populateAuthors` reaches `loadUser()` within the same microtask window. Payload does run afterRead per-doc under `Promise.all`, so this usually holds — but field-level afterRead hooks run first and introduce awaits, so docs can drift into separate ticks and you get partial batching (a few small queries, not one). Not wrong, just less effective than the comments claim.
- **Shared-batch error coupling (behavior change).** Previously each article's populate failed independently. Now a single batched `find` that throws fans the rejection out to every article in that batch. It's swallowed/warned, but it's a real change from per-doc isolation.
- **Minor:** both hooks register a `batchFn` under the same `__articleAuthorLoader` key, but the util reuses the first-registered state and silently ignores the second `batchFn` (harmless only because they're identical). And a util whose correctness hinges on async timing really should ship with unit tests — there are none here.

On the plus side, state lives on `req.context`, which is per-request and shared across the whole `find`, so there's no cross-request leakage — the scope is correct.

### Recommendation

- **Ship the `join` change** after confirming the JSON-LD depth point — that's the idiomatic win and kills the worst N+1.
- **Reconsider the custom loader**, in order of preference:
  1. Expose the handful of public user fields via a read-access rule so native `depth` population works — then the hooks disappear entirely and Payload's own DataLoader batches them.
  2. If you must hand-populate, batch through `req.payloadDataLoader` (or a thin wrapper over the existing `dataloader` dep) instead of a bespoke microtask loop.
  3. If you keep `getRequestLoader`, add unit tests and note in the comments that batching is best-effort.

---

PR #878, opened 2026-07-21
feat(narration): ElevenLabs narration script extractor

### LINKED ISSUE #866: Add "Copy article text" button to feed ElevenLabs for AI voice narration
## Summary

Add a button that extracts an article's text as clean, narration-ready plain text and opens it in an **editable preview** for final tweaks, then copies it to the clipboard to paste into [ElevenLabs](https://elevenlabs.io/) for an AI voice-over. This streamlines producing the audio that the existing narration player plays back.

## Motivation

Articles already support audio narration — `ArticleHero` (`src/heros/ArticleHero/index.tsx`) renders a `NarrationPlayer` (`src/components/NarrationPlayer/index.tsx`) when a `narration` audio track is uploaded on the article. Today, producing that track means an editor manually reconstructs the article text before pasting it into a TTS tool. A one-click "extract the whole article as text" action removes that friction and makes it easy to keep narration in sync with edits.

## Placement (decided)

**Admin-only for now.** A custom UI field/action button on the Articles collection edit view in the Payload admin — generating narration audio is an editorial task, not something writers do inline. See collection config at `src/collections/Articles/index.ts`. A public-facing variant is explicitly out of scope for this issue.

## Proposed behavior — preview & edit, then copy

Rather than a blind one-click copy, the button opens a **preview window** (modal/drawer) so the editor can see and adjust exactly what will be sent to ElevenLabs. A blind copy would hide the parts most likely to need a human eye — legacy-block fallback substitutions, break-tag placement, number/math normalization.

- **Button** (e.g. "Narration text…") opens the preview.
- **Editable text area** prefilled with the extractor output (see "Text extraction"). A large, monospace-ish editor comfortable for reviewing/editing a full article.
- **Copy button** inside the window copies the current (possibly edited) contents to the clipboard via the Clipboard API.
- **Copy feedback uses Payload UI toasts** — `import { toast } from '@payloadcms/ui'` (Payload's re-export of `sonner`). Fire a success toast on copy (e.g. "Narration text copied to clipboard") and an error toast on failure, so feedback matches the rest of the admin rather than a bespoke inline label/icon swap.
- **Edits are ephemeral** — a scratch buffer for this copy only. They are **not persisted** to the article, and reopening re-runs the extractor from the current article content. (Optional niceties: a "Reset to extracted" button; a character/estimated-duration count.)
- Accessible: focus-trapped dialog, `aria-label`s, keyboard operable (incl. Esc to close), and a graceful fallback if the Clipboard API is unavailable (text remains selectable to copy manually, with an error toast).

## Text extraction

Article body is Lexical rich text (`content` field), rendered by `RichText` (`src/components/RichText`). A reusable, independently-testable extractor (e.g. `lexicalToPlainText`) produces the string that prefills the preview editor. It should:

- **Prepend the title and byline** — including author name(s) and the publish date. Data is available on the article (`populatedAuthors`, `publishedAt`).
- Walk the Lexical node tree and concatenate text nodes into readable paragraphs (blank line between blocks).
- Replace non-prose blocks with their accessible description or a legacy fallback (see "Block descriptions" below) — never silently drop them.
- **Footnotes — omit them.** In this publication footnotes are largely URL/citation references, which carry no spoken value: reading a URL aloud is unpleasant, and once the URL is stripped there's little or nothing left to narrate. So the extractor should **drop footnote text and strip the inline `[1]` markers entirely** — no end list, no spoken "footnote N" references. (Article exposes `footnotes` / renders a `FootnoteList` for the visual page; narration simply skips them.)
  - _Deferred:_ if footnotes ever grow substantive non-citation prose, revisit inlining them as spoken asides. Out of scope here.

## Block descriptions (`description` field — accessible, not narration-only)

A voice-over can't read a map, an equation, or an image — but this metadata is **broader than narration**. The same short prose is useful as an **`aria-label` / alt text** on the rendered page, so the field is a general-purpose **`description`** (not a narration-specific name).

- Add a `description` field to **all embeddable blocks** — `Math`, `MediaBlock`, `SocialEmbed`, `SquiggleRule`, `Code`, etc. (`src/blocks/<Name>/config.ts`).
  - _"Writers can provide descriptions of the equation. Descriptions of the image. Descriptions of the map."_
- **Don't duplicate descriptions that already exist.** Some blocks already carry good text — e.g. media **alt text**. Prefer the existing field (media `alt`) and only add a new `description` field where a block has nothing suitable. The extractor picks the best available: explicit `description` → existing alt/caption → legacy fallback.
- **Legacy fallback** for blocks with no description (older articles): `To view this <Block type name>, please refer to the article.` (e.g. "To view this map, please refer to the article.")
  - Note: "view" is imperfect for users relying on narration for visual accessibility, but legacy blocks simply have no description, so pointing to the visual article is the best available fallback until they're backfilled. The preview editor is a good place to catch and hand-fix these before copying.
- The extractor inserts the description (or fallback) inline where the block sits, so the narrated text reads continuously.

## ElevenLabs formatting

Take advantage of ElevenLabs input features to maximize narration quality (not just raw text dump). Investigate and apply what meaningfully helps, e.g.:

- **Pacing via break tags** — insert `<break time="…" />` (or deliberate punctuation/ellipses) between the title, byline, section headings, and paragraphs so the read isn't a run-on.
- **Text normalization** — expand/normalize numbers, symbols, abbreviations, and math notation into spoken form so TTS pronounces them naturally.
- **Structure cues** — clear paragraph breaks and sentence punctuation for natural prosody.
- Keep the output paste-ready; voice/model selection and pronunciation dictionaries are configured in ElevenLabs itself, but the copied text should be shaped to get the best result from them. The preview editor lets an editor hand-add or adjust these before copying.

Document which features are used so the format is intentional and testable.

## Acceptance criteria

- [ ] An accessible, admin-only button opens a preview window containing the extracted article text (title + byline incl. publish date + body) in an **editable** text area.
- [ ] A copy action inside the window copies the current (possibly edited) contents to the clipboard.
- [ ] Copy success and failure are reported via **Payload UI toasts** (`toast` from `@payloadcms/ui`), not a bespoke inline indicator.
- [ ] Edits in the preview are ephemeral — not persisted to the article; reopening re-runs the extractor.
- [ ] Non-prose blocks are replaced by their `description`, existing alt/caption, or a per-type legacy fallback (`To view this <Block type name>, please refer to the article.`) — never silently dropped.
- [ ] A `description` field is added to embeddable blocks that lack existing suitable text, and the same value is wired up to serve `aria-label`/alt on the rendered page (or at least modeled to allow it). Blocks with existing descriptions (media `alt`) are reused, not duplicated.
- [ ] Footnotes are omitted from the copied text and their inline markers are stripped.
- [ ] ElevenLabs-oriented formatting (break tags / normalization / structure) is applied and documented.
- [ ] Lexical → text extraction (block-description substitution, fallbacks, footnote-marker stripping) is covered by a unit test (`src/**/__tests__/`).
- [ ] Works in the admin without breaking SSR.

## Resolved decisions

1. **Placement:** admin-only for now.
2. **Interaction:** button opens an editable preview window; copy happens from inside it. Edits are ephemeral (not persisted).
3. **Copy feedback:** Payload UI toasts (`toast` from `@payloadcms/ui`) for success and error.
4. **Byline:** title + author(s) + publish date included.
5. **Footnotes:** omitted entirely (they're largely URL citations with no spoken value); inline markers stripped.
6. **Block field:** a general-purpose **`description`** field on all embeddable blocks (also usable for `aria-label`/alt), reusing existing fields like media `alt` rather than duplicating.
7. **Legacy fallback:** `To view this <Block type name>, please refer to the article.`
8. **ElevenLabs:** use its input features (break tags, normalization, structure) as needed for high-quality narration.

espdesign:
## Context

Closes #866.

Editors had no way to get an article out of Payload and into ElevenLabs without hand-copying it. This adds a **Narration** tab to Articles with a one-click extractor that turns the Lexical content into a script ready to paste.

The hard part isn't the prose — it's everything that *isn't* prose. Images, formulas, timelines, and embeds either read as gibberish or vanish. Each one now becomes a **parenthetical narrator aside** the voice reads verbatim:

```
(An image: crowds gather outside the courthouse.)
```

This replaced an earlier `<< BLOCK_NAME: detail >>` marker scheme — a TTS engine spells all-caps block names out letter by letter and reads the brackets as punctuation, so the markers ended up in the audio.

### Narration tab

The narration audio upload and the new `ExtractNarrationButton` now live together under one tab; the duplicate sidebar inputs are gone.

### `extractNarrationText`

Walks the Lexical tree and emits a byline (title, authors, published date) followed by the body.

- Horizontal rules, squiggle rules, and footnotes are dropped entirely.
- Media blocks and collages use the caption (preferred) or alt text. Filenames and numeric ids are deliberately omitted — they read terribly aloud and add no narration value.
- Social embeds speak the stored oEmbed title **only** on a successful snapshot fetch. `buildSnapshot` leaves sentinel titles like "No adapter found" behind when it can't resolve a platform and carries stale titles forward on network failures, so any status other than `ok` is treated as no title.
- Raw LaTeX and raw URLs are never read aloud.

### Math blocks: description field + screen reader label

New optional `description` field on the inline and display math blocks — a plain-words name for the formula, e.g. *"the Cauchy–Schwarz inequality"*.

- The voice-over reads it **in place of** the LaTeX. Display math becomes an aside; inline math reads bare so it stays inside the surrounding clause.
- The same string doubles as the formula's accessible name: `role="math"` + `aria-label`, with the rendered MathJax markup `aria-hidden` so a screen reader announces the description instead of reading the markup character by character.
- With no description, the rendered output is byte-for-byte what it was before.

No migration needed — math blocks exist only inside the Lexical editor, so the field lives in the rich-text JSON.

### Timeline narration

A timeline is readable text, not a visual, so it narrates its actual contents rather than collapsing to a stand-in phrase:

```
(A timeline titled "Fall of Rome": August 24, 410: Visigoths sack Rome. Alaric enters the city.
September 4, 476: Romulus Augustulus is deposed.)
```

- `describeTimeline` lives with the block (`src/blocks/Timeline/narration.ts`) — the per-event field shape is the block's business, not the extractor's, which just calls it.
- One line per event, newline-separated rather than bulleted: a literal `-` or `•` is either spelled out or swallowed by a TTS engine, while a line break gives the pause a bullet is there to convey.
- Dates are formatted `en-US`/UTC so an article doesn't sound different depending on the server's locale and timezone.
- Citations and avatars are omitted — a link has nothing to say aloud, and an avatar is exactly the sort of visual this list replaces. An empty event list falls back to `(A timeline.)`.
- `describeVisual` moved to `src/utilities/describeVisual.ts` so block-level narration helpers and the extractor share one phrasing.

### Extractor UI

- An in-memory `Map` cache survives the unmounts caused by switching tabs inside the Payload editor, without a `sessionStorage` layer (which throws in private mode). Manual edits to the script still reset on a hard reload — the UI says so.
- Unpopulated media relationships are batch-fetched in concurrent chunks of 25 to cut request overhead.
- `<break time="3s" />` lines render as badges in the preview so editors can see the SSML pauses.

## Test Plan

- `pnpm test:unit` — covers the parser's fallbacks, nested media relationship shapes, unsupported inline blocks, collage objects, timeline event edge cases (missing/unparseable dates, dateless events, `Date` instances, citation links that must not leak into the script), the math accessibility wrapper, an unsaved article's path-keyed script cache, and the button's clipboard-permission and fetch-failure paths.
- Statement / branch coverage on the new files:

  | File | Statements | Branches |
  | --- | --- | --- |
  | `utilities/describeVisual.ts` | 100% | 100% |
  | `blocks/Math/Component.tsx` | 100% | 100% |
  | `blocks/Timeline/narration.ts` | 100% | 100% |
  | `utilities/extractNarrationText.ts` | 99% | 93% |
  | `Articles/components/ExtractNarrationButton.tsx` | 98% | 84% |

  Two paths are deliberately uncovered: `resolveMediaInfo`'s `if (!mediaVal) return null` is unreachable (both call sites guard on truthiness), and the button's `getServerSnapshot` / `window`-undefined cache key are SSR paths jsdom can't reach.
- Manually: seed with the narration demo (`src/endpoints/seed/features/narration-demo.ts`, updated to exercise the block types above), open an article → **Narration** tab → **Extract**, and confirm the script contains no `<<`/`>>` markers, no raw LaTeX, no URLs, and one line per timeline event.
- Accessibility: with a screen reader, a display formula that has a description announces the description and not the MathJax markup; one without a description behaves as it did before.


src/blocks/Banner/config.ts
espdesign, line 1: Probably don't need both description and text in banner.

src/blocks/SocialEmbed/embeds/BlueskyEmbed/config.ts
espdesign, line 1: Not needed depreciated.

src/blocks/SocialEmbed/embeds/RedditEmbed/config.ts
espdesign, line 1: Not needed depreciated

src/blocks/SocialEmbed/embeds/TikTokEmbed/config.ts
espdesign, line 1: Not needed depreciated

src/blocks/SocialEmbed/embeds/TwitterEmbed/config.ts
espdesign, line 1: Not needed depreciated

src/blocks/SocialEmbed/embeds/YouTubeEmbed/config.ts
espdesign, line 1: Not needed depreciated

src/fields/blockDescription.ts
espdesign, line 1: Maybe reword.

src/utilities/extractNarrationText.ts
espdesign, line 1: Need to review more.

espdesign (commented):
Needs changes

tadjh:
Flagging a shift in direction for this PR so it's documented.

Right now the block fallbacks emit editor markers like `<< TIMELINE: … >>` that a person is expected to find and manually rewrite before the script goes to ElevenLabs. I don't want to ship it that way. Our editors are already tight on time, and asking them to hand-script a readout for every image, timeline, map, embed, etc. is exactly the kind of busywork I want to design *out* of this feature, not into it.

The fallback is the whole point. We should have a sane, automatic default for **every** block type that reads well out of the box — 9 times out of 10 this should be fully hands-off. Editors should rarely need to touch it, and when they do it should be to *improve* on a good default, never to make an unusable placeholder usable.

Concretely: instead of `<< TIMELINE: Title >>` (which TTS either spells out letter-by-letter as "T-I-M-E-L-I-N-E" or reads the brackets aloud), each block should narrate as a natural spoken aside, e.g.:

- `A timeline titled "The Fall of Rome" is shown here.`
- `An image is shown here: crowds gather outside the courthouse.`
- `An embedded YouTube video is shown here.`
- Math and links never read raw LaTeX or URL text aloud.

I've pushed an implementation of this on `claude/block-fallback-elevenlabs-ezhw3i` (commit ff6b1cf) that we can fold in — the parser now auto-describes all block types and the tests are updated to match. Let's go with automatic descriptions for every block rather than manual markers.

---

PR #881, opened 2026-08-02
feat(merch): add a Merch block backed by an hourly Shopify sync

### LINKED ISSUE #880: Add Store page link
## Original request

DiGG now has a store page.

https://store.digitalgroundgame.org/

Add a link to the site so readers can buy our stuff :)

---

## Scope

A **Merch** carousel block that showcases store products as on-site placements (page body, sidebar, or inline in rich text), linking out to the DGG merch pages.

PR #881 builds the block itself — square / full-width layouts, badges, opt-in autoplay, UTM-tagged outbound links. In that PR products were **typed in by hand** per placement, which was **deemed too rough to ship**: every price change, new drop, or sold-out item is a manual CMS edit repeated across every placement, with no signal when a block goes stale.

**So the block ships with syncing.** Products come from a synced collection; hand-typed rows are not the shipping path.

Cart and checkout are explicitly **out of scope for this site** — the reader clicks through to digitalgroundgame.org, which already has its own product pages and cart. On-site checkout here belongs to #648 (Payload ecommerce plugin).

## 1. Sync job

Mirrors the existing `updateRecommendations` job:

- `src/jobs/syncShopifyProducts/{index.ts,logic.ts}`, registered in `payload.config.ts` → `jobs.tasks`.
- `schedule: [{ cron: "0 * * * *", queue: "default" }]`. `jobs.autoRun` already ticks the `default` queue every 5 min, and external runs are already gated on `CRON_SECRET`.
- Admin "run now" endpoint at `/merch/sync`, mirroring `/article-recommendations/run`, for the editor who just pushed a drop.
- `retries: { attempts: 2, backoff: exponential }`, `outputSchema` with the run's counts, and the same `payload.logger` step logging.

**Source API** — the Shopify **Storefront API** (GraphQL), server-side, at the same endpoint dgg-frontpage uses:

```
POST https://{SHOPIFY_STORE_DOMAIN}/api/{SHOPIFY_API_VERSION}/graphql.json
X-Shopify-Storefront-Access-Token: {SHOPIFY_STOREFRONT_ACCESS_TOKEN}
```

Env only: `SHOPIFY_STORE_DOMAIN`, `SHOPIFY_STOREFRONT_ACCESS_TOKEN` (the public storefront token, provided in production), `SHOPIFY_API_VERSION`. Ours is a **server-side** read, unlike dgg-frontpage's `NEXT_PUBLIC_…`.

Two notes on the query:

- It paginates with cursors rather than dgg-frontpage's fixed `first: 24`, and adds `tags`, `collections`, and `featuredImage`'s `width`/`height` (which `next/image` needs for a remote file).
- `collections` requires the `unauthenticated_read_product_listings` scope, and **one unauthorized field fails the whole query** — so a scope error retries once without it. Losing collection filters is survivable; losing the catalogue isn't.

**Sync semantics**

- Upsert by the store's product ID; **never hard-delete** — a delisted product flips to `archived` so blocks referencing it degrade to "not shown".
- Missing credentials skip the run (dev and CI have none), and a failed fetch leaves prior rows rendering. Neither dev nor a Shopify outage blanks a carousel.

## 2. Links point at digitalgroundgame.org, not the Shopify storefront

**The quirk worth getting right.** Shopify is where the catalogue *data* comes from, but readers never land on `store.digitalgroundgame.org` — the DGG site has duplicated the products and cart, and generates its own product pages from the same API.

| Link | Target |
| --- | --- |
| A product card | `https://digitalgroundgame.org/merch/{handle}` — the `[handle]` route in dgg-frontpage |
| "Shop all" | `https://digitalgroundgame.org/merch` |
| Anything | **never** `store.digitalgroundgame.org` |

- `MERCH_SITE_URL` holds the store page **path and all** (`https://digitalgroundgame.org/merch`), so moving the storefront to another path or host is config rather than code. The store domain is used *only* for the API call.
- Product URLs are **derived from the synced handle**, never stored. A handle renamed in Shopify would 404 on the DGG site, so each sync refreshing it is what keeps links alive — and archived products stop being linked at all.
- UTM tagging (`withMerchUtm`) applies; the destination is a sibling DGG property, so it's cross-property attribution.
- **`rel="noopener"`**, not `sponsored nofollow`. DGG is the parent org of The Pragmatic Papers: `sponsored` means paid placement and `nofollow` means "not an endorsement", and neither describes a publication linking to its own parent's store. `noreferrer` is dropped too, so DGG's analytics sees the referrer alongside the campaign params.

## 3. The `merch` collection

`src/collections/Merch/` (slug `merch` — short enough to read well in a query; Payload derives the doc type `Merch` from it).

Fields are named for the **commerce concept, not the vendor**, and each is a leaf value stored as the store reports it:

| Field | Notes |
| --- | --- |
| `externalId` | unique + indexed, the upsert key |
| `source` | `shopify` today; a second storefront adds an option, not a second collection |
| `handle` | unique; the DGG merch URL is derived from it (§2) |
| `title`, `description` | as returned |
| `price`, `compareAtPrice`, `currencyCode` | raw amounts and a code — never a formatted string |
| `availableForSale` | as reported, including unknown |
| `imageUrl`, `imageWidth`, `imageHeight`, `imageAlt` | Shopify CDN URL — see §5 |
| `tags`, `collections` | what filtered blocks narrow on |
| `status` | `active` / `archived` (archived = the store stopped listing it) |
| `lastSyncedAt` | staleness signal in the admin list view |
| **editorial** `featured`, `badgeOverride`, `hidden`, `sortOrder` | ours, never overwritten by a sync |

`collections` is a plain list of Shopify collection handles — the store's own product groupings, what a shopper reads as a category — and has **nothing to do with Payload collections**, despite the word. It keeps Shopify's name for two reasons: every other field here uses Shopify's vocabulary, so translating one of them would leave a single mismatch for anyone comparing a row against the Storefront response; and Shopify separately uses "category" for its product taxonomy, which is a different thing we may want that name for. The overload is documented at the collection, at the field, in the admin description, and at the query that filters on it.

Two decisions behind that shape:

- **The sync transcribes, it doesn't interpret.** No formatted prices, no deciding that a `"0.0"` compare-at means "not on sale", no coercing unknown availability to false. Those are display choices, and baking one into the database means a re-sync to change your mind. The block owns them.
- **No raw payload column.** This table is a *derived cache* — every row is reproducible from the store within the hour — so a raw copy has no archival value: a field we skipped becomes a column and the next run backfills it. Raw payloads earn their place for irreproducible sources (webhook deliveries, audit trails), which this isn't.

Flattening `priceRange.minVariantPrice.amount` to `price` isn't reshaping the data — `priceRange` and `collections { nodes }` are GraphQL transport, not Shopify's data model.

Access: public read for active products, admin-only write; the job writes with `overrideAccess`. Commerce fields render read-only in the admin. No drafts/versions.

## 4. Block sources from the collection

`source` select on the block:

- **`all`** — every active, non-hidden synced product; ordered by `sortOrder` / title / newest, with a `limit`.
- **`filtered`** — narrow by `collection` (a Shopify collection handle), `tag`, or the `featured` flag; or hand-pick exact products (and their order) via a `hasMany` relationship to `merch`.

`getMerchProducts()` in `src/blocks/Merch/products.ts` is the single normalizer: the source branch, URL derivation, price formatting (`Intl.NumberFormat`, as in dgg-frontpage's `money()`), and badge derivation all live there, and the carousel component is unchanged. The block renderer is async.

## 5. Product images — remote URLs

The `cdn.shopify.com` URL is stored and rendered directly, with a `remotePatterns` entry in `next.config.ts`. No ingestion into `Media`, no duplicated assets.

`ImageMedia` handles it: `getMediaUrl()` passes absolute URLs through, and synced products are shaped into a Media-like object (`url` / `width` / `height` / `alt` / `mimeType`) at the normalizer boundary — hence pulling image dimensions in the query. A remote file has no resize variants, so the block's per-layout `sizes` hints do the responsive work.

## 6. Cache invalidation

- The product query is cached under a `merch` tag, the way `getCachedGlobal` tags globals.
- The sync revalidates that tag **only when a run actually created, updated, or archived something** — the hourly no-op costs nothing.
- Editorial edits revalidate the same tag from the collection's `afterChange` / `afterDelete` hooks.

## Testing

- **Unit** — Storefront → row mapper, the query builder (including the collections-less variant), pagination, change detection, URL derivation, price formatting, badge derivation, the Media-shape adapter, `source` resolution, and the block's rendering and link attributes.
- **Integration** (Testcontainers, `fetch` mocked) — idempotent upserts, archive-not-delete, renamed handles, a failed fetch leaving prior data intact, editorial fields surviving a sync, and archived products restoring when the store lists them again.
- **E2E** — the existing merch spec, updated for the DGG links and `rel`.

## Reference

The shopping experience on digitalgroundgame.org: https://github.com/digitalgroundgame/dgg-frontpage — `src/app/merch/page.tsx`, `src/app/merch/[handle]/page.tsx`, `src/components/merch/merch-store.tsx`.

tadjh:
## Context

https://pr-881.pragmaticpapers.com/home

Closes #880 (superset). The issue asks for a link to the DGG store; this adds a reusable **Merch** block that showcases store products on-site, backed by an hourly sync so the products stay true on their own.

The block originally stored its products **by hand** — image, title, price, badge, URL, typed per placement. That was deemed too rough to ship: every price change, new drop, or sold-out item became a manual CMS edit repeated in every placement, with no signal when a block went stale. So the hand-typed rows came out and the block now stores a *query* against a synced catalogue.

One quirk drives much of the design: **Shopify is only the data source.** Readers must land on the DGG site's own product pages (`MERCH_SITE_URL/{handle}` — the products and cart are duplicated there, in [dgg-frontpage](https://github.com/digitalgroundgame/dgg-frontpage)), never on the storefront host. Every link is derived from the handle at render time, so a rename in Shopify is picked up by the next sync rather than leaving a dead link behind.

### What landed

**`merch` collection** (`src/collections/Merch/`) — the local mirror of the catalogue.

- Commerce fields are stored as the store reports them — `price` is a decimal string, not a formatted one; `availableForSale` is a boolean, not a badge. Presentation is the block's job, so a currency or copy change never means a re-sync. Each field is a leaf value named for the commerce concept rather than the vendor, and there is no raw payload alongside: this is a derived cache, so a field we skipped can be added as a column and backfilled by the next run.
- `create` is denied outright — rows come from the sync, and a hand-made row has no real store ID to upsert against, so the next run would archive it immediately. That also removes Payload's "Create New" button and empty state.
- Products are **never hard-deleted**. One that unpublishes flips to `archived`, so a block pointing at it degrades to "not shown" rather than a dangling reference.
- A collapsible of **editorial fields** (`featured`, `hidden`, `badgeOverride`, `sortOrder`) is ours and survives every sync.
- `collections` keeps Shopify's name for the store's own product groupings — documented at the top of the collection, since it has nothing to do with Payload collections.

**Hourly sync job** (`src/jobs/syncShopifyProducts/`) — mirrors the existing `updateRecommendations` task.

- `TaskConfig` on `0 * * * *`, 2 retries with exponential backoff, counts in its `outputSchema`.
- Storefront GraphQL with cursor pagination (`MAX_PAGES` bounded), upserting on the store's product ID.
- Degrades rather than blanks: missing credentials log a warning and no-op (dev/CI have none), a failed fetch leaves prior rows intact, and a store whose token lacks the collections scope retries without that field instead of failing the run.
- Revalidates the `merch` cache tag **only** when a run actually changed something.
- A **"Sync from Shopify now"** button above the admin list (`POST /api/merch/sync`, admin-only) for the editor who just pushed a drop. It reports honestly when credentials are absent, rather than claiming success over an empty catalogue.

**Block sources from the collection** (`src/blocks/Merch/`) — `source: all | filtered`, with filters on Shopify collection handle, tag, featured-only, and hand-picked products (which display in the arranged order), plus `orderBy` and `limit`. `getMerchProducts()` stays the single normalizer that everything renders through: it derives the link, formats the price via `Intl.NumberFormat` from the synced currency, derives the "Sold Out" badge (an editor's `badgeOverride` wins), and shapes the remote CDN image into the `Media` contract `ImageMedia` expects. The carousel, layouts, autoplay, and controls are unchanged from the original block.

**Outbound links** — still tagged with campaign params (`utm_source=pragmaticpapers`, `utm_medium=merch_block`, `utm_content` per placement), since the destination site owns the analytics and this one ships none. `rel` is now **`noopener`**, not `sponsored nofollow noopener noreferrer`: DGG is the parent org of The Pragmatic Papers, so the links are neither paid placements nor unendorsed, and leaving `noreferrer` off lets DGG's analytics see where a click came from.

**Where the store lives is config, not content** — every link comes from `MERCH_SITE_URL`. With it unset the block renders no products and no "Shop all" rather than guessing at a URL. There is no per-block override and no address baked into the repo.

**Also in here**

- `RichText` is now **server-only** (`src/components/RichText/index.tsx`), with `internalDocToHref` split out for the client callers. The merch converter reaches the Payload config, and two client components were pulling the whole converter map into the browser bundle through it; the Form block now server-renders its rich text and passes the elements down.
- `next.config.ts` gains a `cdn.shopify.com` remote pattern — product shots are rendered from the CDN, not copied into Media. A `Cell` component puts the thumbnail in the admin list so a row is recognisable at a glance.
- New env vars, documented in `.env.example`: `SHOPIFY_STORE_DOMAIN` (scheme tolerated), `SHOPIFY_STOREFRONT_ACCESS_TOKEN`, `SHOPIFY_API_VERSION`, `MERCH_SITE_URL`. The e2e workflow and `docker-compose.e2e.yml` set `MERCH_SITE_URL`.
- Seeds create `merch` docs and reference them (`src/endpoints/seed/merch.ts`), keeping the same titles, prices, badges, and images so the existing visual baselines stay valid — only hrefs and `rel` changed, and neither is visual.
- Migration `src/migrations/20260812_230124_add_merch_collection.{ts,json}`, replayed from an empty database.

## Test Plan

Automated — **88 unit / RTL tests** and **8 integration tests**:

- `logic.test.ts` (28) — env reading and domain normalisation (a scheme in `SHOPIFY_STORE_DOMAIN` is tolerated, since that's what took the first production run down), endpoint building, pagination, the Shopify→row mapper, change detection, and the collections-scope fallback.
- `products.test.ts` (21) — price formatting, badge derivation and override, the Media shape for a remote image, link derivation, and the query each `source`/filter combination builds.
- `Component.test.tsx` (16) — empty state, headings, prices, badges, `rel="noopener"`, campaign params, the store button appearing only when configured, and the square vs full-width slide classes.
- `MerchCarousel.test.tsx` (7) — autoplay gating on the prop and on reduced-motion, arrows hidden when nothing scrolls, dots from snap points.
- `urls.test.ts` (5) / `utm.test.ts` (5) — link derivation from config alone, and campaign tagging that leaves editor-supplied params and unparseable URLs untouched.
- `SyncNowButton.test.tsx` (4) / `ProductThumbnailCell.test.tsx` (2) — the admin affordances.
- `tests/integration/syncShopifyProducts.test.ts` (8, `fetch` mocked) — pagination walks past one page, upserts are idempotent, a vanished product archives rather than deletes, a renamed handle updates the link, a failed fetch leaves prior data intact, editorial fields survive a sync, and a no-op run doesn't revalidate.
- `tests/e2e/merch.spec.ts` — the seeded home-page carousel: six products, tagged links pointing at the DGG merch pages and never the storefront host, `rel="noopener"`, badge and price visible, both arrows present, plus the visual baseline.
- `pnpm lint:ci`, `pnpm check-types`, `pnpm test:unit`, `pnpm test:integration`, and `pnpm build` all pass; migrations were replayed from an empty Postgres with `migrate:fresh`.

Manual:

1. Set `SHOPIFY_STORE_DOMAIN`, `SHOPIFY_STOREFRONT_ACCESS_TOKEN`, `SHOPIFY_API_VERSION`, and `MERCH_SITE_URL` in the environment. The token needs the `unauthenticated_read_product_listings` scope.
2. In the admin, open **Store → Merch** and click **Sync from Shopify now**. Products appear with thumbnails, prices, and a `lastSyncedAt`; without credentials the button says so instead of reporting success.
3. Add a **Merch** block to a Page's Content tab, or embed one in a Content block's rich text. Leave `source` on *All synced products*, then try *Filtered selection* with a collection handle, a tag, featured-only, and a hand-picked set.
4. Toggle `layout` between **Square (sidebar)** and **Full width (page)** and confirm the carousel density and controls (arrows / dots) change; optionally toggle autoplay and confirm it pauses on hover and focus and is suppressed under reduced motion.
5. On the front end, confirm cards link to `MERCH_SITE_URL/{handle}` in a new tab with `utm_*` params attached, and that no link points at the storefront host. Unset `MERCH_SITE_URL` and confirm the block renders nothing rather than a dead link.
6. Mark a product `hidden` or set a `badgeOverride`, run the sync again, and confirm the editorial values survive.


evan6seven:
If you link to the store please use https://digitalgroundgame.org/merch

espdesign:
nitpick, background on lightmode maybe bump it a bit more dark? 

image

hard to see the white on offwhite imo.

src/collections/Merch/index.ts
```diff
@@ -0,0 +1,300 @@
...
+        return Response.json({ jobId: job.id, configured: readShopifyEnv() !== null, result })
+      },
+    },
+  ],
+  fields: [
+    {
+      name: "title",
+      type: "text",
+      required: true,
+      admin: { readOnly: true },
```
espdesign: I was able to change the title in the admin panel on staging.

espdesign: I think it should completely hide the fields we would never edit on pragmatic papers, maybe only show in edit these? 

Image

tadjh: Good catch hardened the schema so that those extra fields are no longer editable.

tadjh:
image

@espdesign Darkened the background for light shirts on grey.

PR #889, opened 2026-08-09
fix(queries): eliminate N+1 volume lookups and resolve relationships natively

### LINKED ISSUE #861: Fix N+1 Query on Articles

espdesign:
Primary goal to resolve https://github.com/digitalgroundgame/pragmatic-papers/issues/861

While investigating the **N+1 Query** performance issue (Sentry: `PRAGMATIC-PAPERS-3`), we found that after the access-control refactor we could lean on Payload's native relationship resolution instead of per-document `afterRead` hooks. Doing that required opening `users.read` — so this PR also locks the private user fields down at the field level.

> [!NOTE]
> #891 has been merged into this branch, so its security fix and review follow-ups are included below rather than stacked on top.

## 1. Fix N+1 volume lookups (the headline change)

The `populateVolume` afterRead hook ran a **per-article reverse `payload.find`** on volumes (`where: { articles: { contains: doc.id } }`) for every article document returned by any query — the homepage was issuing ~9 of these (`select distinct volumes … left join volumes_rels … where articles_id = $N`) plus matching count queries, on top of ~47 total SQL queries per render.

- **Removed `populateVolume` from the Articles `afterRead` hooks** and deleted the hook plus the virtual `populatedVolume` field (and regenerated `payload-types.ts`).
- **Added a shared cached `queryVolumesForArticles(articleIds)`** helper in `src/utilities/queries.ts` that resolves the containing volumes for a batch of article IDs with a single `where: { articles: { in: [...] } }` query. It is keyed on normalized (deduped, sorted) IDs rather than array identity, so two callers in the same request share one query instead of each hitting the DB.
- **Adapted the one real consumer** — the article page JSON-LD — to pass the volume explicitly: `buildArticleJsonLd(article, path, volume)` now takes the volume as an argument instead of reading the virtual field. `VolumeArticle.tsx` uses its existing `volume` prop for the fallback excerpt.
- **De-duplicated** the identical local `queryVolumesForArticles` helpers in the topics and authors pages to use the shared one.

**Verified in Sentry on the preview deploy (pr-889), measured at `671b784`:**

| metric | BEFORE `670f11e` | AFTER `671b784` |
|---|---|---|
| SQL queries on `GET /` | 47 | 16 |
| spans | 111 | 39–40 |
| per-article volume lookups (`… where articles_id = $N`) | 8 | **0** |
| users queries | 5 | 3 |

All remaining homepage queries are batched `IN (…)` lookups (topics, authors, media, articles). Later commits adjusted query depths (see §3), so exact counts may differ slightly from the table; the per-article volume lookups stay at 0.

## 2. Access control

- **`readUsers` policy** — new collection-level read policy in `policies.ts`: anyone may read staff/author users, users may read themselves, staff may read all users. This is what makes native relationship population work for public article bylines.
- **`selfOrAdminFieldLevel` field policy** — new field-level read policy in `fields.ts` restricting `email` and `roles` to the user themselves or an admin/chief-editor. Opening `users.read` is required for native population; widening PII access is not.
- **`STAFF_ROLES` extracted** in `access/roles.ts` so `isStaff` and the `readUsers` where-clause derive the staff set from one list rather than two copies that happened to agree.
- **`queryUserBySlug` now enforces access** — it was the only query in `queries.ts` omitting `overrideAccess`, which the Local API defaults to `true`. That skipped `readUsers` entirely (a duplicated `roles: { in: [...] }` where-clause was the only thing filtering non-staff records) and skipped field access on `email`/`roles` along with it. It now passes `overrideAccess: draft` like every sibling query, and the hand-rolled filter is gone.

### Security fix: `populateAuthors` leaked `email` and `roles` publicly (#891)

The hook fetched users with `overrideAccess: true` — which skips field-level access — and assigned the raw docs straight into `doc.authors`. Collection `afterRead` hooks run **after** Payload's field sanitization (`collections/operations/find.js`: "afterRead - Fields" precedes "afterRead - Collection"), so whatever the hook wrote was serialized to the client verbatim. Two ways to reach it:

1. **Anonymously, at depth 0** — articles are publicly readable and REST accepts a `depth` param, so `GET /api/articles?depth=0` returned every author's `email` and `roles`.
2. **At any depth** — when a related user fails the `readUsers` check, Payload deliberately leaves the bare ID in place (`relationshipPopulationPromise.js`: "ids are visible regardless of access controls"). That tripped the hook's `every(typeof === "object")` guard, so it refetched **all** authors with `overrideAccess: true` and overwrote the already-sanitized ones too.

This defeated exactly the field-level protection this PR adds. It is ultimately fixed by **deleting the hook** (below); the fix is first *contained* in an intermediate commit (an explicit public-safe `select`, backfilling only entries Payload left as bare IDs) so the security fix stays reviewable on its own, independent of the removal.

## 3. Schema & hook cleanups

- **Removed virtual fields** `populatedAuthors` and `populatedNarrator` from the Articles config.
- **Removed `populateAuthors` entirely.** The hook ran on **every** article read to serve two callers that can each resolve authors themselves:
  - **The search reindex** fetches at `depth: 0` (`plugin-search/dist/utilities/generateReindexHandler.js:103`), so it genuinely sees bare IDs. `beforeSync` now resolves names itself with a single batched, name-only query — `BeforeSync` receives `payload` and may return a `Promise` (`plugin-search/dist/types.d.ts:10-18`). Failures fall back to already-populated names, so a transient error degrades one index entry rather than the sync. Same one-query-per-article cost as the hook, just moved to the caller.
  - **`feed.volumes`**, fixed by the depth change below.

  Everything else — article pages, volume pages, the homepage, `feed.articles`, the newsletter, email preview — already ran at depth ≥ 2 and never invoked the hook. Deleting it removes the *class* of bug above (an `afterRead` hook writing `overrideAccess` data past field sanitization) rather than constraining it.

  **One behavior change worth naming:** an author who fails the public read check is now left as a bare ID instead of being backfilled, so their byline would not render. That only affects a contributor demoted after publication — `filterOptions` blocks selecting a non-contributor in the first place. There are none today, and how past contributors should be handled is deliberately deferred.
- **Removed `populateNarrator`** — the narrator relationship now resolves natively under `narration.narrator`.

### Query depths

Depths became load-bearing once relationships resolve natively, so they are written out explicitly rather than relying on Payload's `defaultDepth` — author images and the narrator line break silently if it changes.

- `queryVolumeBySlug` → `depth: 3` (volume → articles → authors → profileImage).
- `queryArticleBySlug` / `queryPageBySlug` → explicit `depth: 2` (article → authors → profileImage; layout blocks → articles → authors).
- `feed.volumes` → `depth: 2`. Its `volume.articles` came back as objects but `article.authors` as bare IDs, so the byline only rendered because `populateAuthors` backfilled them. The route is `force-static` with a 20-volume limit, so the extra joins are paid at build time.

## 4. Frontend & search plugin adaptation

- **Frontend/utilities** — all frontend components, the structured-data helper, the RSS generator, and email templates (`structuredData.ts`, `generateRssFeed.ts`, `VolumeArticle.tsx`, etc.) now map and filter authors from the standard `authors` relationship field.
- **Search plugin** — `beforeSync` in `src/plugins/index.ts` indexes author names from the standard `authors` field via the new `resolveAuthorNames` helper.
- **Extract Narration Button** — resolves authors dynamically for the client-side generator by fetching user documents from `/api/users`. Dropped a fallback that returned `{ name: String(id) }` for an unresolved author — it would have been read aloud as the byline ("By 42") whenever the fetch failed.

## 5. Additional fixes

- **Collection tile** — restored the default `imagePosition` to `above` in `CollectionTile.tsx`.
- **Volumes `generateStaticParams`** — limit restored from 100 back to 1000 (unrelated leftover; with `pagination: false` Payload still applies `limit`, so volumes past the first 100 stopped being prerendered).

## 6. Tests

- **Unit** — `resolveAuthorNames` (batching, order preservation, name-only select, unresolvable authors, error fallback), `queryVolumesForArticles` batching, `queryUserBySlug` (which had no coverage — the `overrideAccess` flag is the kind of thing that regresses silently), and `selfOrAdminFieldLevel` / `readUsers`. Updated `structuredData`, `ArticleHero`, `ExtractNarrationButton`, `formatAuthors`, and `extractNarrationText` to the native-author model.
- **Integration** — `email`/`roles` withheld from anonymous users and from non-admin staff reading another user; article authors populate for anonymous readers without `email`/`roles`; depth 0 returns plain IDs; a user who fails the read check is left as a bare ID without affecting co-authors; members and anonymous readers can read staff profiles.

Worth recording: `filterOptions` is enforced on **write**, so a `member` can never be *selected* as an author — Payload rejects it with a `ValidationError`. Demotion after publication is the only way an article ends up pointing at a non-public user, and the integration test models it that way.

## What is deliberately not here

How contributor visibility should be modelled — so that a writer can lose write access without their byline disappearing — is a separate design question, out of scope for this PR. Nothing here depends on the outcome.


src/app/(frontend)/volumes/[slug]/page.tsx
```diff
@@ -25,19 +25,15 @@ export async function generateStaticParams(): Promise<{ slug: string | null | un
   const volumes = await payload.find({
     collection: "volumes",
     draft: false,
-    limit: 1000,
+    limit: 100,
```
Comment (espdesign) [inline src/app/(frontend)/volumes/[slug]/page.tsx]: Likely unintended change: volumes/[slug]/

page.tsx generateStaticParams - limit: 1000

limit: 100. This is unrelated to the N+1/

access-control work and looks like a leftover from testing. If there are >100 volumes, this will silently cap which volume pages get statically generated. Worth reverting unless intentional.

src/access/policies.ts
```diff
@@ -22,6 +22,31 @@ export const isSelfOrAdmin: Access = ({ req: { user } }) => {
...
+    constraints.push({
+      id: { equals: user.id },
+    })
+  }
+
+  return {
+    or: constraints,
+  }
+}
+
```
espdesign, line 49: Privilege scope check: readUsers grants isStaff(user) (narrator/writer/editor/chief-editor/ admin) unrestricted read on all user docs, and selforStaffFieldLevel uses the same check for email/roles. That means any narrator or writer can now read every member's email - previously only admins or the user themselves could. Might be intentional given team size, but flagging since it's a real PII-access expansion buried in a perf-focused PR.

src/utilities/queries.ts
```diff
@@ -65,7 +65,7 @@ export const queryVolumeBySlug = cache(async (slug: string) => {
         equals: slug,
       },
     },
-    depth: 2,
+    depth: 3,
```
espdesign, line 62: queryVolumeBySlug depth 2 → 3: given the whole PR's goal is cutting query count, worth confirming this doesn't meaningfully offset the win (more joins per depth level). A comment noting why depth 3 is needed (author images) would help future readers.

espdesign (commented):
Testing needs work

**1. Write the missing tests against `dev` (pre-refactor) first**
Check out `dev`, and write tests for the *behavior*, not the implementation, of the three gaps:
- `queryVolumesForArticles`-equivalent (the old duplicated helpers in `topics/[slug]/page.tsx` and `authors/[slug]/page.tsx`) — test "given article IDs, returns the volumes containing them," batched.
- The old `populateVolume`/`populateAuthors`/`populateNarrator` hooks — test what they attach to a doc for a few realistic cases (article with volume, article with no volume, article with multiple authors).
- `ExtractNarrationButton`'s author-resolution path — test what byline text gets generated for a given article.

These become your **behavioral baseline**. They should pass on `dev` before you touch anything.

**2. Rebase/cherry-pick those tests onto the PR branch, unchanged**
Don't rewrite the test's assertions for the new shape yet — first see what breaks. If a test fails only because of *shape* (e.g. expects `article.populatedVolume.title` but the new code returns `volume.title` as a separate argument), that's expected and fine — update the test to call the new API surface, but keep the same input→output expectations.

If a test fails because the **actual output value changed** (not just the access pattern), that's a real regression the PR introduced and needs fixing before merge, not a test update.

**3. Add new tests for genuinely new logic**
The `.filter((a): a is User => typeof a === "object")` pattern and `fetchAuthorsMap`'s chunking/error-fallback are new — write fresh unit tests for these directly on the PR branch, since there's no old behavior to characterize.

**4. Concretely, in order:**
1. `git checkout dev`
2. Add unit tests for `queryVolumesForArticles` (old duplicated versions), old `populateVolume`/`populateAuthors`/`populateNarrator` hooks, and `extractNarrationText`'s author-resolution behavior — commit these on a small branch off `dev`.
3. `git checkout n1-query-access-control-fix` (the PR branch), then cherry-pick or manually port that test commit over.
4. Run the ported tests — update only the call-shape (arguments/imports), not the expected values.
5. Anything that fails on real value mismatch → flag as a PR bug, not a test-update.
6. Add new tests for `fetchAuthorsMap` and the repeated type-guard filter logic, since these have no `dev`-side equivalent.
7. Push the test additions as a review suggestion or a small follow-up commit on the PR branch itself (contributors usually appreciate this more than just a comment saying "please add tests").

This gives you actual proof the refactor is behavior-preserving where it should be, rather than just visually reading old vs. new code and hoping the filter logic is equivalent everywhere it's duplicated.

src/collections/Articles/hooks/populateAuthors.ts
```diff
@@ -20,37 +20,20 @@ export const populateAuthors: CollectionAfterReadHook<Article> = async ({
...
       overrideAccess: true,
-      depth: 1,
     })
 
     // Preserve original author order
     const authorMap = new Map(authorDocs.map((userDoc) => [userDoc.id, userDoc]))
-    const populatedAuthors = authorIds
+    doc.authors = authorIds
       .map((id) => authorMap.get(id))
       .filter((userDoc): userDoc is User => Boolean(userDoc))
```
tadjh: **Blocker (security): this now injects unsanitized user docs — including `email` and `roles` — into publicly readable article responses.**

The chain (each step verified against `payload@3.84.x` source in `node_modules`):

1. `overrideAccess: true` skips field-level read access entirely (`fields/hooks/afterRead/promise.js`: `canReadField = overrideAccess ? true : await field.access.read(...)`), so `authorDocs` here contain `email` and `roles`.
2. These raw docs are assigned to `doc.authors`, and collection `afterRead` hooks run **after** Payload's field-access sanitization of the doc (`collections/operations/find.js`: "afterRead - Fields" precedes "afterRead - Collection"). Whatever this hook writes is serialized verbatim into REST responses.
3. Articles are publicly readable (`isPublishedOrStaff`), and REST accepts a `depth` param — so anonymous `GET /api/articles?depth=0` returns every author's `email` and `roles`.

It also triggers at **any** depth, not just depth 0: when a related user fails the `readUsers` check (e.g. an author later demoted to `member`), Payload's populator intentionally leaves the bare ID in place (`relationshipPopulationPromise.js`: "ids are visible regardless of access controls"). That makes the `every(typeof === "object")` guard on line 13 fail, and the hook then refetches **all** authors with `overrideAccess: true` and overwrites the sanitized ones too.

This defeats exactly the protection this PR adds via `selfOrStaffFieldLevel` on `email`/`roles`. The old hook had the same `overrideAccess: true` fetch but only ever copied a sanitized subset into `populatedAuthors` — that was the load-bearing part.

Suggested fix: constrain the fetch to the public-safe fields, e.g.

```ts
const { docs: authorDocs } = await payload.find({
  collection: "users",
  where: { id: { in: authorIds } },
  depth: 0,
  overrideAccess: true,
  select: { name: true, slug: true },
})
```

(`beforeSync` only needs `.name`; the volumes RSS feed only needs `.name` too.) I'd avoid fixing this by gating the hook to a search-only context flag — `feed.volumes` fetches at `depth: 1` (articles land at depth 0) and relies on this hook for author names, and it would silently lose them.

Worth adding an integration test: anonymous read of `articles` at `depth: 0` must not expose `authors[].email` — the new tests cover the `users` collection surface but not this article-side path.

---

src/access/policies.ts
```diff
@@ -22,6 +22,31 @@ export const isSelfOrAdmin: Access = ({ req: { user } }) => {
...
+  ]
+
+  if (user?.id) {
+    constraints.push({
+      id: { equals: user.id },
+    })
+  }
+
+  return {
+    or: constraints,
```
tadjh, line 46: **Verified safe, for the record:** I checked whether the `roles` query constraint returned here could itself be rejected for anonymous users now that `roles` has field-level read access (`selfOrStaffFieldLevel`, which returns `false` for anonymous). It can't: Payload validates only the *incoming* `where` against field read access, not the access-result constraint (`collections/operations/find.js` calls `validateQueryPaths({ where })` with the raw incoming `where`, while the access result is merged separately into `fullWhere`). So anonymous reads of staff users work, and anonymous users still can't craft their own `where[roles]` queries against `/api/users` (fails closed). The added integration tests genuinely cover this.

**One behavior change worth a deliberate decision:** any user whose roles are later reduced to just `member` (e.g. a departed author) becomes invisible to anonymous readers. On natively populated paths (article pages, JSON-LD, RSS at depth ≥ native population) their byline silently disappears, while hook-populated paths (depth-0/1) still show them — inconsistent, and data-dependent on production roles. Worth confirming no current article author is `member`-only in prod before merging, and deciding what the policy for past authors should be.

---

src/app/(frontend)/volumes/[slug]/page.tsx
Comment (tadjh) [inline src/app/(frontend)/volumes/[slug]/page.tsx]: **Worth noting:** this drops the `generateStaticParams` cap from 1000 to 100, which looks unrelated to the PR's goal. With `pagination: false` Payload still applies `limit` (`find.js`: `sanitizedLimit = limit ?? ...` feeds `db.find`), so volumes beyond the first 100 stop being prerendered. Not breaking — no `dynamicParams = false` anywhere under `src/app/(frontend)` (verified by grep), so they'd render on demand — but if it's intentional it deserves a mention in the PR description, and if not, it's an accidental behavior change.

---

src/utilities/queries.ts
```diff
@@ -85,6 +85,28 @@ export const queryArticleBySlug = cache(async (slug: string) => {
...
+    where: {
+      articles: {
+        in: articleIds,
+      },
+    },
+    depth: 0,
+  })
+
+  return docs
+})
```
tadjh: Two small notes here:

1. **PR description vs code:** the description says `queryPageBySlug`/`queryVolumeBySlug` were bumped to depth 3, but only `queryVolumeBySlug` changed (commit `fab8802` touches one line). `queryPageBySlug` and `queryArticleBySlug` run at Payload's `defaultDepth` of 2 (verified: `payload.config.ts` doesn't set `defaultDepth`; `config/defaults.js` defaults it to 2). That happens to be exactly enough today — page → article → authors, and article → narration → narrator both fit in 2 levels — but it's an implicit dependency: `ArticleHero`'s narrator line and author profile images now break silently if anyone sets a lower explicit depth here. A `depth: 2` written out explicitly (or a comment) would protect that.

2. **`React.cache` keyed by array identity:** `queryVolumesForArticles` memoizes on the `articleIds` array *reference*, so two callers building their own `[id, …]` arrays in the same request will each hit the DB. Same behavior as the per-page helpers this replaces, so no regression — just don't count on it as a cross-component dedupe.

---

tadjh (commented):
## Code review — consequence-focused pass (self-hosted Coolify + Cloudflare, Payload 3, Next 15)

This PR touches two of the highest-blast-radius surfaces in the repo: **Payload access control** (`users.read` opened up from `isSelfOrAdmin` to `readUsers`, new field-level restrictions on `email`/`roles`) and **afterRead hooks** (three populate hooks removed, `populateAuthors` rewritten to mutate `doc.authors` in place). No migrations needed (the removed fields were `virtual`, no columns — verified no diff under `src/migrations/`), no env/header/Docker/Sentry/storage changes.

**Verdict: one security blocker, otherwise the design is sound and well-tested.** The N+1 fix itself is correct and the Sentry before/after numbers back it up.

### Findings

1. **Blocker — `populateAuthors.ts` leaks `email` + `roles` into public article responses.** The rewritten hook assigns raw user docs fetched with `overrideAccess: true` (which skips field-level access) directly into `doc.authors`, and collection `afterRead` hooks run *after* Payload's field sanitization — so the injected values go to the client verbatim. Reachable anonymously via `GET /api/articles?depth=0`, and at *any* depth for articles with an author who fails the public read check (Payload leaves the bare ID, tripping the hook's refetch-all path). This defeats the exact field-level protection the PR adds. Fix is small: `select: { name: true, slug: true }` on the hook's find. Full trace in the inline comment.

2. **Worth noting — demoted authors vanish inconsistently.** A user whose roles drop to `member`-only disappears from natively populated bylines/JSON-LD/RSS but still appears via hook-populated paths. Data-dependent; worth checking prod before merge (inline comment on `policies.ts`).

3. **Worth noting — description drift on depth.** Only `queryVolumeBySlug` went to depth 3; `queryPageBySlug`/`queryArticleBySlug` rely implicitly on Payload's `defaultDepth: 2`, which is now load-bearing for narrator + author images (inline comment on `queries.ts`).

4. **Worth noting — `volumes` `generateStaticParams` limit silently dropped 1000 → 100** (inline comment; looks unrelated to the PR goal).

### Verified safe (so you don't have to re-check)

- The `roles` constraint returned by `readUsers` is **not** rejected for anonymous users: `validateQueryPaths` only validates the incoming `where`, not access-result constraints (verified in `payload@3.84.x` `collections/operations/find.js`). Conversely, anonymous users can't query `where[roles]` themselves — fails closed.
- Redefining `email` on the `auth: true` collection is safe: `mergeBaseFields` deep-merges with Payload's base email field, so `required`, `unique`, the lowercase-trim hook, and validation are all retained — no login or Drizzle schema drift.
- `selfOrStaffFieldLevel` receives `id: doc.id` in the field afterRead traversal, so it works in list reads as well as `findByID` (members keep seeing their own email in list contexts, staff short-circuit).
- Hidden auth fields (`hash`, `salt`, `sessions`, reset tokens) stay hidden on anonymous `/api/users` — `showHiddenFields` isn't reachable via REST.
- No frontend query filters on `users.roles` with `overrideAccess: false` (`queryAuthors`/`queryUserBySlug` use the local-API default `overrideAccess: true`), so the new field restriction breaks no existing query.
- `feed.volumes` (depth 1) and the search plugin's `beforeSync` (depth 0) still get author names — via the rewritten hook, which is why finding #1 should be fixed with `select` rather than by removing/gating the hook.

Couldn't verify from the repo: live Cloudflare/Coolify config (not in version control) — but nothing in this diff touches headers, caching directives, or proxy assumptions, so no manual check needed there.

---

tadjh (dismissed):
looking good, unless you have more tests to add

tadjh (approved):

PR #890, opened 2026-08-10
fix(access): stop author field leak, add author role for past contributors

tadjh:
Stacked on top of #889 — **base branch is `n1-query-access-control-fix`, not `dev`**. Merge #889 first and this retargets to `dev` on its own.

Everything here is review follow-up to #889. Splitting it out so #889 stays @espdesign's, and the security fix and the role change can be reviewed on their own terms.

## 1. Blocker: `populateAuthors` leaked `email` and `roles` publicly

The rewritten hook fetched users with `overrideAccess: true` — which skips field-level access — and assigned the raw docs into `doc.authors`. Collection `afterRead` hooks run **after** Payload's field sanitization (`collections/operations/find.js`: "afterRead - Fields" precedes "afterRead - Collection"), so whatever the hook wrote was serialized verbatim.

Reachable anonymously via `GET /api/articles?depth=0`, and at *any* depth when a related user failed the read check: Payload leaves the bare ID in place, which tripped the hook's `every(typeof === "object")` guard and made it refetch **all** authors, overwriting the sanitized ones too. This defeated exactly the protection #889 added via field-level access on `email`/`roles`.

**Ultimately fixed by deleting the hook** (§3). It was first contained with an explicit public-safe `select` and a merge that only backfilled bare IDs, and that commit is kept in the history so the fix is reviewable independently of the removal.

## 2. `email`/`roles` narrowed back to admin-or-self

#889 opened these to `isStaff`, so any writer or narrator could read every member's email — previously only admins or the user themselves could. Opening `users.read` is genuinely required for native relationship population; widening PII access is not. `selfOrStaffFieldLevel` → `selfOrAdminFieldLevel`.

## 3. Removed `populateAuthors` entirely

The hook ran on every article read to serve two callers that can each resolve authors themselves:

- **Search reindex** fetches at `depth: 0` (`plugin-search/dist/utilities/generateReindexHandler.js:103`), so it genuinely sees bare IDs. `beforeSync` now resolves names itself with a batched, name-only query — `BeforeSync` receives `payload` and may return a `Promise`.
- **`feed.volumes`** ran at `depth: 1`, so `volume.articles` were objects but `article.authors` were IDs. Now `depth: 2`.
- Everything else — `feed.articles`, the newsletter, email preview, article and volume pages — already ran at depth 2 or higher.

Removing it deletes the class of bug in §1 rather than containing it.

## 4. New `author` role — decoupling credit from permission

`roles` was doing two unrelated jobs: expressing what an account may *do*, and implying whether a person is publicly *credited*. Since `readUsers` and the `authors` `filterOptions` both key off current roles, offboarding a writer by dropping them to `member` silently stripped their byline and profile page from articles they had already published. Authorship is a fact about a published work, not a permission.

- **`author` grants nothing** (`isStaff` deliberately excludes it) and only keeps a contributor publicly readable and byline-eligible.
- **`ensureAuthorRole`** adds it alongside any contributor role, so stepping down is one action: remove `writer`, `author` remains. A documented convention would decay immediately and fail silently — the byline just disappears. Collection `beforeChange` runs before the field hooks enforcing `roles`' admin-only update access, so this is not an escalation path.
- **`CONTRIBUTOR_ROLES` / `BYLINE_ROLES` / `PUBLIC_PROFILE_ROLES`** replace the same role list hand-rolled across five call sites — the drift that made this easy to miss.
- **Migrations**: one adds `'author'` to the `enum_users_roles` enum (with the Drizzle snapshot, without which the pending-migrations check sees permanent drift); a second backfills it to every current contributor and to anyone already credited on an article who holds no contributor role. Separate files because Postgres will not use a newly added enum value in the transaction that added it.

## 5. Smaller review findings from #889

- Restored the `volumes` `generateStaticParams` limit from 100 back to 1000 (unrelated leftover).
- `queryArticleBySlug`/`queryPageBySlug` write out `depth: 2` explicitly rather than relying on Payload's `defaultDepth`, which became load-bearing once relationships resolved natively — author images and the narrator line break silently if it changes. (#889's description said these went to depth 3; only `queryVolumeBySlug` did.)
- `queryVolumesForArticles` is keyed on normalized (deduped, sorted) IDs rather than array identity, so two callers in the same request share one query.
- Removed a narration byline fallback that emitted an unresolved author's numeric ID as their name — it would have been read aloud.

## Tests

- **Unit**: `resolveAuthorNames` (batching, ordering, name-only select, error fallback), `ensureAuthorRole`, `queryVolumesForArticles`, and the role constants. 575 passing.
- **Integration**: `email`/`roles` withheld from anonymous users, non-admin staff, and public article responses; a member-only user left as a bare ID rather than backfilled; the `author` role keeping a byline intact across offboarding while granting no read access.

One test taught us something worth recording: `filterOptions` is enforced on **write**, so a `member` can never be *selected* as an author — Payload rejects it with a `ValidationError`. Demotion after publication is the only way an article ends up pointing at a non-public user.

## Before merging

The backfill migration grants `author` from existing data, so no manual step is needed — but it's worth eyeballing who ends up with the role in prod, since it's derived from current `users_roles` plus `articles_rels`.


tadjh:
Closing — superseded by #891 and out of scope for the N+1 work.

The security findings (the `email`/`roles` leak in `populateAuthors`, the PII narrowing back to admin-or-self) and the small review findings from #889 all live in #891, which is self-contained.

What this PR was additionally trying to solve — keeping a past contributor's byline and profile page alive after they lose the `writer` role — is deferred. There are no past contributors today, and it isn't N+1 scope. Three designs were explored and none are worth carrying now:

- an `author` role: a role that grants nothing, needing a hook to infer credit from permissions, plus an irreversible Postgres enum value
- a `publicProfile`/`contributor` field: same denormalization, needs a backfill and can drift
- a join field on Users: `traverseFields.js:294-307` shows joins populate by default (`defaultLimit ?? 10`, and `0` means unlimited), so it would attach a subquery to every user read — self-defeating for a perf PR

Worth recording for whenever this comes back: #891's `select` fix already makes bylines durable, because the hook backfills authors Payload leaves as bare IDs. The only remaining gap is that such a contributor's `/authors/[slug]` page 404s and they drop off the `/authors` index — both gateable on 'has published articles' per page, with no durable marker anywhere.

PR #891, opened 2026-08-10
fix(access): stop populateAuthors leaking email/roles into public article responses

tadjh:
Stacked on #889 — **base branch is `n1-query-access-control-fix`, not `dev`**. Retargets to `dev` on its own once #889 merges.

Split out of the larger review follow-up so the security fix isn't blocked on an unrelated design discussion about how contributor visibility should be modelled.

## Blocker: `populateAuthors` leaked `email` and `roles` publicly

The hook fetched users with `overrideAccess: true` — which skips field-level access — and assigned the raw docs straight into `doc.authors`. Collection `afterRead` hooks run **after** Payload's field sanitization (`collections/operations/find.js`: "afterRead - Fields" precedes "afterRead - Collection"), so whatever the hook wrote was serialized to the client verbatim.

Two ways to reach it:

1. **Anonymously, at depth 0** — articles are publicly readable and REST accepts a `depth` param, so `GET /api/articles?depth=0` returned every author's `email` and `roles`.
2. **At any depth** — when a related user fails the `readUsers` check, Payload deliberately leaves the bare ID in place (`relationshipPopulationPromise.js`: "ids are visible regardless of access controls"). That tripped the hook's `every(typeof === "object")` guard, so it refetched **all** authors with `overrideAccess: true` and overwrote the already-sanitized ones too.

This defeated exactly the field-level protection #889 added.

**Ultimately fixed by deleting the hook** — see "Removing `populateAuthors`" below. It is first *contained* in the opening commit (an explicit public-safe `select`, and backfilling only entries Payload left as bare IDs) so the security fix stays reviewable on its own, independent of the removal.

## `email`/`roles` narrowed back to admin-or-self

#889 opened these to `isStaff`, so any writer or narrator could read every member's email — previously only admins or the user themselves could. Opening `users.read` is genuinely required for native relationship population; widening PII access is not. `selfOrStaffFieldLevel` → `selfOrAdminFieldLevel`.

## Removing `populateAuthors`

The hook ran on **every** article read to serve two callers that can each resolve authors themselves:

- **The search reindex** fetches at `depth: 0` (`plugin-search/dist/utilities/generateReindexHandler.js:103`), so it genuinely sees bare IDs. `beforeSync` now resolves names itself with a single batched, name-only query — `BeforeSync` receives `payload` and may return a `Promise` (`plugin-search/dist/types.d.ts:10-18`). Failures fall back to already-populated names, so a transient error degrades one index entry rather than the sync. Same one-query-per-article cost as the hook, just moved to the caller; the reindex is a per-document batch loop by design.
- **`feed.volumes`**, fixed by the depth change below.

Everything else — article pages, volume pages, the homepage, `feed.articles`, the newsletter, email preview — already ran at depth ≥ 2 and never invoked the hook.

Deleting it removes the *class* of bug above (an `afterRead` hook writing `overrideAccess` data past field sanitization) rather than constraining it.

**One behavior change worth naming:** an author who fails the public read check is now left as a bare ID instead of being backfilled, so their byline would not render. That only affects a contributor demoted after publication — `filterOptions` blocks selecting a non-contributor in the first place. There are none today, and how past contributors should be handled is deliberately deferred.

## Smaller findings from the #889 review

- Restored the `volumes` `generateStaticParams` limit from 100 back to 1000 (unrelated leftover; with `pagination: false` Payload still applies `limit`, so volumes past the first 100 stopped being prerendered).
- `queryArticleBySlug`/`queryPageBySlug` now write out `depth: 2` explicitly rather than relying on Payload's `defaultDepth`, which became load-bearing once relationships resolved natively — author images and the narrator line break silently if it changes. (#889's description says these went to depth 3; only `queryVolumeBySlug` did.)
- `queryVolumesForArticles` is keyed on normalized (deduped, sorted) IDs rather than array identity, so two callers in the same request share one query instead of each hitting the DB.
- `feed.volumes` moved from `depth: 1` to `depth: 2`. Its `volume.articles` came back as objects but `article.authors` as bare IDs, so the byline only rendered because `populateAuthors` backfilled them. Resolving natively removes one of the two remaining reasons that hook exists. The route is `force-static` with a 20-volume limit, so the extra joins are paid at build time.
- Removed a fallback in `ExtractNarrationButton` that returned `{ name: String(id) }` for an unresolved author — it would have been read aloud as the byline ("By 42") whenever the `/api/users` fetch failed.

## Tests

- **Unit**: `resolveAuthorNames` (batching, order preservation, name-only select, unresolvable authors, error fallback) and `queryVolumesForArticles` batching.
- **Integration**: `email`/`roles` withheld from anonymous users and from non-admin staff reading another user; article authors populate for anonymous readers without `email`/`roles`; depth 0 returns plain IDs; a user who fails the read check is left as a bare ID without affecting co-authors.

Worth recording: `filterOptions` is enforced on **write**, so a `member` can never be *selected* as an author — Payload rejects it with a `ValidationError`. Demotion after publication is the only way an article ends up pointing at a non-public user, and the integration test models it that way.

## What is deliberately not here

How contributor visibility should be modelled — so that a writer can lose write access without their byline disappearing — is a separate design question, out of scope for this PR. Nothing in this PR depends on the outcome.


espdesign (approved):
LGTM

Issue #916, opened 2026-08-17
Restore next/link client-side navigation once Cloudflare caches RSC payloads correctly


tadjh:
Tracking issue for re-enabling `next/link`. Supersedes #589, which attempted this and was closed unmerged — a closed PR can't track the work, and the code has moved since.

## Background

**#575 (merged Apr 6)** replaced every `next/link` with plain anchor tags. The stated reason:

> While cloudflare seems to be caching our pages, the RSC query param is rarely the same when going from one page to the next, which makes most internal navigation cold/MISS. Since our origin server is slow, uncached client-side navigations take several seconds.

So the problem is **not** that Cloudflare ignores `Vary` and serves HTML where flight is expected. It's the opposite: Next appends a `?_rsc=` hash to RSC requests, that hash varies by originating route, and Cloudflare keys on the full query string — so a given target page gets a separate cache entry per route you navigate from. Prefetches almost always MISS and fall through to a slow origin. Full page loads, by contrast, have one stable cache key and hit the edge.

**#589 (closed unmerged, Jun 1)** tried to restore links alongside Cloudflare Rewrite Rules, citing [Cloudflare's "serve tailored content" recipe](https://developers.cloudflare.com/cache/advanced-configuration/serve-tailored-content/#example-caching-nextjs-rsc-payloads), which covers caching Next.js RSC payloads via a custom cache key. It went stale and now conflicts — its paths are all `apps/pragmatic-papers/…`, predating the monorepo flatten (#519).

## Current state

`HoverPrefetchLink` (`src/components/Link/HoverPrefetchLink.tsx`) still carries the full `next/link` implementation commented out — the `Link` import, `usePathname`, `useState`, and the `prefetch` / `onMouseEnter` / `data-active` props — and renders a bare anchor. `CMSLink`, `CMSLink2`, and `LinkButton` are plain anchors too. `eslint.config.mjs` disables `@next/next/no-html-link-for-pages` to permit this.

## Work

**1. Cloudflare (do first — this is the actual blocker).** Apply the Cache Rule from the CF recipe: a custom cache key that includes the `RSC` request header and normalizes away the volatile `_rsc` query param, so HTML and flight variants each get **one** stable entry per URL rather than one per (target, origin-route) pair.

[...]

PR #936, opened 2026-08-23
fix: restore active state in header navigation

### LINKED ISSUE #617: Bug: Header menu missing active page hint
The header menu used to display a hint on what the active page is. It should be added back.
Sidebar could also display active page hint.

Investigate `data-active` not being set for the `navigation-menu-item` elements.

Ashish-Minj1290:
Closes #617

Restores active navigation state in the MegaMenu using Base UI's `active` prop and adds unit tests for exact, nested, root, trailing-slash, and external-link route matching.

## CI follow-up
The first Playwright run failed in the pre-existing `author-card.spec.ts` test because its locator matched two responsive copies of the same seeded author card (one hidden, one visible), causing a Playwright strict-mode violation. The application behavior under this PR was not the source of that failure: 26/27 E2E tests passed and the core CI workflow was green.

This branch now scopes that E2E locator to the visible card only, preserving the existing assertions while avoiding the hidden duplicate.

CI should provide the definitive validation for the updated head.

### DIFF
--- a/src/components/MegaMenu/index.tsx
+++ b/src/components/MegaMenu/index.tsx
@@ -1,3 +1,5 @@
+"use client"
+
 import {
   NavigationMenu,
   NavigationMenuItem,
@@ -5,13 +7,26 @@ import {
   NavigationMenuList,
 } from "@/components/ui/navigation-menu"
 import { type MenuField } from "@/payload-types"
+import { getLinkFieldUrl } from "@/utilities/getLinkFieldUrl"
+import { usePathname } from "next/navigation"
 import { CMSLink } from "../Link/CMSLink2"
 
 interface MegaMenuProps {
   menu?: MenuField
 }
 
+function isActivePath(pathname: string, url: string | null): boolean {
+  if (!url?.startsWith("/")) return false
+
+  const normalizedUrl = url === "/" ? url : url.replace(/\/+$/, "")
+  if (normalizedUrl === "/") return pathname === "/"
+
+  return pathname === normalizedUrl || pathname.startsWith(`${normalizedUrl}/`)
+}
+
 export function MegaMenu({ menu }: MegaMenuProps): React.ReactNode {
+  const pathname = usePathname()
+
   if (!menu) return null
   return (
     <div className="my-2 hidden w-full justify-center md:flex">
@@ -26,11 +41,18 @@ export function MegaMenu({ menu }: MegaMenuProps): React.ReactNode {
         <NavigationMenuLink>Link</NavigationMenuLink>
       </NavigationMenuContent>
     </NavigationMenuItem> */}
-          {menu.map((item) => (
-            <NavigationMenuItem key={item.id}>
-              <NavigationMenuLink className="py-1" render={<CMSLink link={item.link} />} />
-            </NavigationMenuItem>
-          ))}
+          {menu.map((item) => {
+            const url = getLinkFieldUrl(item.link)
+            return (
+              <NavigationMenuItem key={item.id}>
+                <NavigationMenuLink
+                  active={isActivePath(pathname, url)}
+                  className="py-1"
+                  render={<CMSLink link={item.link} />}
+                />
+              </NavigationMenuItem>
+            )
+          })}
         </NavigationMenuList>
       </NavigationMenu>
     </div>
--- a/src/components/MegaMenu/index.tsx
+++ b/src/components/MegaMenu/index.tsx
@@ -10,18 +10,10 @@ import { type MenuField } from "@/payload-types"
 import { getLinkFieldUrl } from "@/utilities/getLinkFieldUrl"
 import { usePathname } from "next/navigation"
 import { CMSLink } from "../Link/CMSLink2"
+import { isActivePath } from "./isActivePath"
 
 interface MegaMenuProps {
   menu?: MenuField
-}
-
-function isActivePath(pathname: string, url: string | null): boolean {
-  if (!url?.startsWith("/")) return false
-
-  const normalizedUrl = url === "/" ? url : url.replace(/\/+$/, "")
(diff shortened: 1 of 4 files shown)


src/components/MegaMenu/index.tsx
```diff
@@ -28,7 +29,11 @@ export function MegaMenu({ menu }: MegaMenuProps): React.ReactNode {
     </NavigationMenuItem> */}
           {menu.map((item) => (
             <NavigationMenuItem key={item.id}>
-              <NavigationMenuLink className="py-1" render={<CMSLink link={item.link} />} />
+              <MegaMenuLink
+                href={getLinkFieldUrl(item.link)}
```
tadjh: **Recommend: skip items without a URL.** When `getLinkFieldUrl` returns `null` (the referenced doc was deleted, or its cached `reference.value` is no longer populated, see #970), `CMSLink` renders nothing, but Base UI's `NavigationMenu.Link` still renders its fallback `<a>` with the `p-2 py-1` padding and hover styles, so the header shows an empty, highlightable box (already the case on `dev`). Resolve the URL once here and `return null` for the item when it's missing. That drops the empty element, and lets `MegaMenuLink` take `href: string` instead of `string | null`.

src/components/Menu/index.tsx
```diff
@@ -66,13 +68,17 @@ export const Menu: React.FC<MenuProps> = ({ menu, className, layout, slot, ...pr
...
               <Slot>
-                <CMSLink
-                  link={link}
-                  className={cn(
-                    "block w-full text-left",
-                    isStacked &&
-                      "data-[active=true]:bg-muted px-4 py-3 data-[active=true]:font-semibold",
-                  )}
+                <MenuLink
+                  href={getLinkFieldUrl(link)}
```
tadjh: **Recommend: skip items without a URL.** Same case as in `MegaMenu`: when `getLinkFieldUrl` returns `null`, `CMSLink` renders nothing and `useRender` falls back to an empty `<a></a>`, inside an empty `SheetClose` button in the sidebar. Resolve the URL once and `return null` for the `<li>` when it's missing, so `MenuLink` can take `href: string`.

src/utilities/isActivePath.ts
```diff
@@ -0,0 +1,8 @@
+export function isActivePath(pathname: string, url: string | null): boolean {
+  if (!url?.startsWith("/")) return false
+
+  const normalizedUrl = url === "/" ? url : url.replace(/\/+$/, "")
```
tadjh: **Recommend: compare the path portion only.** A custom menu URL like `/topics?sort=new` or `/about#team` is never active, because `usePathname()` has no query or hash but `url` does. Stripping everything from the first `?` or `#` before normalizing fixes it, plus a test case for each.

src/components/MegaMenu/MegaMenuLink.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { NavigationMenuLink } from "@/components/ui/navigation-menu"
+import { usePathname } from "next/navigation"
+import { isActivePath } from "@/utilities/isActivePath"
+
+type MegaMenuLinkProps = Omit<React.ComponentProps<typeof NavigationMenuLink>, "href"> & {
+  href: string | null
+}
+
+export function MegaMenuLink({ href, ...props }: MegaMenuLinkProps): React.ReactNode {
```
tadjh: **Recommend: add a test.** `MenuLink` has RTL coverage but `MegaMenuLink` is only covered by a manual browser check. A `MegaMenu` test (it's a plain sync component) covering the active/inactive marking and the skipped null-URL item would cover both this leaf and the fix above.

tadjh (commented):
Review at `f3a42cbc`. CI is green and the approach holds up: `MegaMenu` and `Menu` stay server components, only the `href` string crosses to the client leaves, and the page source carries no `navItems` or reference documents. No infra surfaces touched (headers, env, access, hooks, migrations, caching, auth).

Inline comments below cover two small follow-ups, which I'll push to this branch:
1. A nav item whose target can't be resolved renders an empty anchor (a blank, hoverable box in the header). This mostly predates the PR but is cheap to close here.
2. `isActivePath` never matches a custom URL with a `?query` or `#hash`, and `MegaMenuLink` has no test.

Also noted on #916: when `next/link` is restored there, the old `isCurrent` / `data-active` lines in `HoverPrefetchLink` should stay deleted, since this PR now owns the active state.

src/components/MegaMenu/index.tsx
tadjh: Applied in 560b4efd: `MegaMenu` resolves the URL once and skips the item when it's `null`, and `MegaMenuLink` now takes `href: string`.

src/components/Menu/index.tsx
tadjh: Applied in 560b4efd: `Menu` skips the `<li>` when the URL is `null`, and `MenuLink` now takes `href: string`. Covered by the new `Menu.test.tsx`.

src/utilities/isActivePath.ts
tadjh: Applied in 3744219c: compares only the path before the first `?` or `#`, with test cases for a query, a hash, and `/?ref=nav` on the home route. `url` narrows to `string` since both menus now skip missing URLs.

src/components/MegaMenu/MegaMenuLink.tsx
tadjh: Applied in 560b4efd: `MegaMenu.test.tsx` renders the real `MegaMenu` with a mocked `usePathname` and checks `aria-current` / `data-active` on the active and inactive links, plus the skipped null-URL item.

tadjh (approved):

Issue #660, opened 2026-05-15
Unit tests for pure utility functions


tadjh:
> **Status note (updated 2026-08-12):** Partially covered. `src/utilities/__tests__/`
> now has `formatAuthors.test.ts` and `getInitials.test.ts` from this list (plus
> `toRoman`, the reference example, and out-of-scope `listmonk`, `turnstile`,
> `extractNarrationText`, `structuredData`). **Still untested from the scope below:**
> `formatDateTime`, `deepMerge`, `getLinkFieldUrl`, `toKebabCase`, `getSeparator`,
> `canUseDOM`, `authorSlug`, `isAutosave`, `prettify`, `mergeOpenGraph`,
> `generatePreviewPath`, `utils`.

---

## Goal
Add unit tests for every pure function in `src/utilities/` — functions with no I/O, no Payload calls, and no Next.js dependencies.

## Reference
`src/utilities/__tests__/toRoman.test.ts` is the existing example to follow for style.

## Scope
- `formatAuthors.ts` — oxford-comma author list formatting
- `formatDateTime.ts` — date-fns timestamp formatting
- `deepMerge.ts` — recursive object merge
- `getLinkFieldUrl.ts` — extracts URL from internal/external link field
- `toKebabCase.ts` — string conversion
- `getInitials.ts` — name initials extraction
- `getSeparator.ts` — list separator helper
- `canUseDOM.ts` — DOM environment detection
- `authorSlug.ts` — author slug generation
- `isAutosave.ts` — autosave detection
- `prettify.ts` — display formatting
- `mergeOpenGraph.ts` — OG metadata merge
- `generatePreviewPath.ts` — preview URL path construction
- `utils.ts` — general helpers

## Acceptance Criteria
- Each file has a corresponding `__tests__/*.test.ts` colocated in `src/utilities/`
- Tests cover happy path, edge cases (empty input, nulls), and boundary values
- `pnpm test:unit` passes with no new failures

SalmaAbdelrhmanMostafaMahmoud:
Can I take this issue if it's still available please ?

tadjh:
[...]

PR #937, opened 2026-08-23
test: add coverage for pure utility helpers

Ashish-Minj1290:
## Summary
- add unit coverage for `deepMerge` and `isObject`
- add unit coverage for `toKebabCase`
- add timezone-safe unit coverage for `formatDateTime`

Part of #660

## Coverage
The tests exercise nested merges, source overrides, immutability, camelCase/whitespace conversion, empty input, date formatting, and the empty-timestamp fallback.

No production code is changed. CI should provide the definitive validation.

### DIFF
--- a/src/utilities/__tests__/deepMerge.test.ts
+++ b/src/utilities/__tests__/deepMerge.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from "vitest"
+import deepMerge, { isObject } from "../deepMerge"
+
+describe("deepMerge", () => {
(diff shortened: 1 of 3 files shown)


tadjh:
Superseded by #958, which carries over your `deepMerge` test commit with authorship intact. Thanks, @Ashish-Minj1290! Building on your tests, adding null cases turned up two real bugs in `deepMerge`, now fixed there. The `formatDateTime` tests were dropped because #922 had already added a suite on `dev`, and `toKebabCase` turned out to be unused, so it was deleted instead of tested.

PR #948, opened 2026-09-16
fix: buttons no pointer-cursor in tailwind v4

### LINKED ISSUE #902: Buttons site-wide have no pointer cursor since Tailwind v4
## What's wrong

Every `<button>` on the site shows the default arrow cursor on hover instead of a pointer. Tailwind v3's preflight included:

```css
button,
[role="button"] {
  cursor: pointer;
}
```

v4 dropped it in favour of the browser default (`cursor: default`) — it's a listed breaking change in the [v4 upgrade guide](https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor). We're on `tailwindcss@^4.2.4` and nothing in the codebase restores it:

- `src/components/ui/button.tsx` — no `cursor-*` in the base `cva` string or any variant.
- `src/app/(frontend)/globals.css` — no `cursor` rule anywhere.
- `grep -rn "cursor-pointer" src` returns exactly two places: `src/blocks/MediaBlock/LightboxMediaBlock.tsx:25` and the carousel controls (added in #881 after the carousel arrows were reported as feeling inert).

So this is not a carousel problem — it's every button, and it's been quietly true since the v4 upgrade. Links are unaffected (`<a>` keeps its own cursor).

## Expected

Hovering an enabled button shows `cursor: pointer`. Disabled buttons keep the default cursor.

## Suggested fix

Restore the v3 behaviour once, in the base layer, rather than sprinkling `cursor-pointer` across components — that's the upgrade guide's own recommendation and it can't drift:

```css
@layer base {
  button:not(:disabled),
  [role="button"]:not(:disabled) {
    cursor: pointer;
  }
}
```

Notes for whoever picks this up:

- `buttonVariants` already carries `disabled:pointer-events-none`, so a disabled `Button` never receives hover in the first place; the `:not(:disabled)` guard matters for plain `<button disabled>` elements outside that component.
- Once the base rule lands, the two ad-hoc `cursor-pointer` usages above become redundant and can be dropped in the same PR.
- Payload's admin panel has its own styles and isn't affected either way.

Not urgent, but it's a small, self-contained fix with a visible effect on every interactive control.

cielestrial:
## Context

- Add cursor-pointer to buttons in tailwind's base layer directive
- Remove cursor-pointer class from buttons' class list

## Test Plan

Hamburger button, search button, and close button should now show expected cursor.


tadjh:
Closes #902

PR #949, opened 2026-09-19
fix: author revalidation and seeding

### LINKED ISSUE #769: Fix Author Revalidation and Seeding
[\[141257.699\] INFO (payload59310) Re.txt](https://github.com/user-attachments/files/28614368/141257.699.INFO.payload59310.Re.txt)

We revalidate `/author` after every single user created during seeding. Fix this and revalidate `/authors` once after all users created.

cielestrial:
Target Fix (author revalidation and seeding):
- Add default value to `createUser` context parameter
- Update `createUsers` to use disable context
- Revalidate `/authors` path in `seed/index` after users have been created
- Update the definition of `createUser` in `making-seeds`

Tests:
- Add test for `createUsers` to check that every create carries the disable context and its value is `true`
- Update `createUser` test to expect the disable context

Warnings:
- Fix `PayloadAdminBar` warning for client side functions not ending in "Action"
- Fix deprecated field in `vscode/settings.json`

Misc:
- Floating the idea of using `organizeImports` in `vscode/settings.json`


src/endpoints/seed/users.ts
```diff
@@ -13,11 +13,13 @@ export interface SeededUsers {
...
   Omit<Partial<User>, keyof RequiredDataFromCollection<User>>
 
+export type UserContext = Record<string, unknown> & { disableRevalidate: boolean }
+
 export async function createUser(
   payload: Payload,
   data: UserData,
   label: string,
-  context?: Record<string, unknown>,
+  context: UserContext = { disableRevalidate: false },
```
tadjh: The default `{ disableRevalidate: false }` works the same as passing no context, because `revalidateUser` only checks `context.disableRevalidate`. It also makes `disableRevalidate` required for every caller, which is why `createUser.test.ts` had to change. Keeping the type (it documents the flag) but making the flag optional and dropping the default gives the same behaviour with less churn.

_This suggestion spanned a removed line, so GitHub couldn't apply it. It's been split into two single-line suggestions on lines 16 and 22 below._

src/endpoints/seed/users.ts
```diff
@@ -266,12 +271,15 @@ export const createUsers = async (payload: Payload, media: Media[]): Promise<See
...
+    context,
   )
 
   const writers: User[] = []
   for (let i = 0; i < WRITER_DATA.length; i++) {
     const writerData = generateWriterData(i, media)
-    const writer = await createUser(payload, writerData, `writer${i + 1}`)
+    const writer = await createUser(payload, writerData, `writer${i + 1}`, {
+      disableRevalidate: true,
+    })
```
tadjh: Writers get a new inline object while everyone else gets the shared `context`. Reuse it here too:

```suggestion
    const writer = await createUser(payload, writerData, `writer${i + 1}`, context)
```

src/endpoints/seed/index.ts
```diff
@@ -102,6 +103,7 @@ export const seed = async (
         ctx.writers = writers
         ctx.narrator = narrator
         validateWriters([writers[0]!, writers[1]!])
+        revalidatePath("/authors")
```
tadjh: `revalidateUser` used to revalidate `/authors/<slug>` for each user as well as the index. Now nothing touches the author pages. One pattern call covers all of them:

```suggestion
        revalidatePath("/authors")
        revalidatePath("/authors/[slug]", "page")
```

src/endpoints/seed/__tests__/createUser.test.ts
```diff
@@ -21,6 +20,8 @@ const userData = {
 
 const mockUser = { id: 1, ...userData }
 
+const defaultContext: UserContext = { disableRevalidate: false }
+
```
tadjh: If the default is dropped (see the `users.ts` suggestion), this helper isn't needed and the two assertions below go back to `context: undefined`.

```suggestion

```

src/endpoints/seed/__tests__/createUser.test.ts
```diff
@@ -35,14 +36,14 @@ describe("createUser", () => {
     expect(mockCreate).toHaveBeenCalledWith({
       collection: "users",
       data: userData,
-      context: undefined,
+      context: defaultContext,
```
tadjh:
```suggestion
      context: undefined,
```

src/endpoints/seed/__tests__/createUser.test.ts
```diff
@@ -67,7 +68,7 @@ describe("createUser", () => {
         roles: userData.roles,
         slug: userData.slug,
       },
-      context: undefined,
+      context: defaultContext,
```
tadjh:
```suggestion
      context: undefined,
```

src/endpoints/seed/__tests__/createUsers.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    expect(mockCreate).toHaveBeenCalled()
+    for (const call of mockCreate.mock.calls) {
+      const options = call[0]
+      expect(options).toEqual(
+        expect.objectContaining({ context: expect.objectContaining({ disableRevalidate: true }) }),
+      )
+    }
+  })
+})
```
tadjh: A few ways to tighten this test:
- `mockResolvedValueOnce` means only the first create returns a user and the other 25 get `undefined`. It passes, but by accident. `mockImplementation` returns a user for every call.
- `toHaveBeenCalled()` would still pass if the loop stopped early. Asserting the exact count (derived from the result, so there's no magic number) checks that every user went through `createUser`.
- `userData` / `mockUser` are only used for the single mock, so they can go.

```suggestion
describe("createUsers", () => {
  it("disables per-user revalidation on every create", async () => {
    mockCreate.mockImplementation(async ({ data }) => ({ id: 1, ...data }))

    const { writers } = await createUsers(mockPayload, [])

    expect(mockCreate).toHaveBeenCalledTimes(writers.length + 4)
    for (const [options] of mockCreate.mock.calls) {
      expect(options).toEqual(expect.objectContaining({ context: { disableRevalidate: true } }))
    }
  })
})
```

src/endpoints/seed/making-seeds.md
```diff
@@ -336,7 +336,12 @@ async function createArticle(payload: Payload, options: CreateArticleOptions): P
...
 ```typescript
-async function createUser(payload: Payload, data: UserData, label: string): Promise<User> {
+async function createUser(
+  payload: Payload,
+  data: UserData,
+  label: string,
+  context?: UserContext,
+): Promise<User> {
   try {
     return await payload.create({ collection: "users", data })
```
tadjh: The example now takes `context` but never uses it. Pass it through here:

```suggestion
    return await payload.create({ collection: "users", data, context })
```

The retry below is outside this diff so it can't take a suggestion, but it also needs `context,` in its `payload.create` call. While you're there, `role` → `roles` (a leftover from before this PR) to match `users.ts`.

tadjh (commented):
Thanks for picking up #769. Batching the revalidation is the right fix, and running it once after the users step works because the seed runs inside a Next route handler.

**`.vscode/settings.json`**: I pushed 8fd17e06 to restore this file to `dev`, and moved the discussion to #969. The `js/ts.tsdk.path` rename is correct for current VS Code, but Cursor may not know the new key yet, so we'll keep both keys for a release. Organize imports is welcome, just without commented-out settings in a shared config file.

The inline suggestions below were checked together locally: seed tests pass, and ESLint, Prettier and `check-types` are clean. They depend on each other (the `users.ts` default and the `createUser.test.ts` assertions), so use **Add suggestion to batch** and commit them as one.

src/endpoints/seed/users.ts
```diff
@@ -13,11 +13,13 @@ export interface SeededUsers {
 type UserData = RequiredDataFromCollection<User> &
   Omit<Partial<User>, keyof RequiredDataFromCollection<User>>
 
+export type UserContext = Record<string, unknown> & { disableRevalidate: boolean }
```
tadjh: Replaces the earlier 16–22 suggestion, which GitHub couldn't apply because the range crossed a removed line. Part 1 of 2: make the flag optional.

```suggestion
export type UserContext = Record<string, unknown> & { disableRevalidate?: boolean }
```

src/endpoints/seed/users.ts
tadjh: Part 2 of 2: drop the default, which works the same as passing no context. Batch this with part 1 and the three `createUser.test.ts` suggestions, or those tests will fail.

```suggestion
  context?: UserContext,
```

tadjh (dismissed):
LGTM!

tadjh (approved):

